PR #8, opened 2018-03-04
Google AdChoices icon layout centering

talklittle:


chauduyphanvu1:
Hi, thanks for submitting the Pull Request. Since AdMob only lets you position the AdChoices icon in [one of the four corners](https://github.com/mopub/mopub-android-mediation/blob/d75bab1bd27b116a5788e8aca14850e5ad5c23f8/AdMob/com/mopub/nativeads/GooglePlayServicesNative.java#L462-L465), I'm going to close the PR.

talklittle:
Hi @chauduyphanvu1 that's not always true. See https://developers.mopub.com/docs/mediation/networks/google/

> The Google Mobile Ads SDK supports a different setting to allow publishers to specify the location of the AdChoices placement, and it addresses the layouts with margins noted above. However, your account must be enabled by AdMob for this feature. If you are affected by this margin issue and placing AdChoices in one of the four corners of the native ad view is not sufficient, please contact your account manager to request this feature.

And find usages of `GooglePlayServicesAdRenderer.VIEW_BINDER_KEY_AD_CHOICES_ICON_CONTAINER`

talklittle:
Note that this pull request only affects that exceptional case that I quoted, because in the normal case where you put the icon in one of the 4 corners, `staticNativeViewHolder.mAdChoicesIconContainer` is null, so the LayoutParams specified in this pull request never take effect.

chauduyphanvu1:
You are correct. However, the positioning should take place in a publisher's implementation, rather than in the adapter even if it only takes effect under some circumstances (for code maintenance purposes). MoPub has a code snippet for whitelisted publishers who would like to position the AdChoices icon elsewhere.

PR #12, opened 2018-03-12
Certify Unity Ads 2.2.0

chauduyphanvu1:
Also passing the existing placement ID to Unity's isReady() check.


UnityAds/com/mopub/mobileads/UnityRouter.java
```diff
@@ -94,7 +94,6 @@ public void onUnityAdsClick(String placementId) {
 
         // @Override
         public void onUnityAdsPlacementStateChanged(String placementId, UnityAds.PlacementState oldState, UnityAds.PlacementState newState) {
-
```
rsatrasala: You removed bunch of empty lines. Was that intensional?

chauduyphanvu1: Yep, these empty lines shouldn't be there.

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
```diff
@@ -71,7 +71,7 @@ protected void loadWithSdkInitialized(@NonNull Activity activity,
         mLauncherActivity = activity;
 
         UnityRouter.addListener(sPlacementId, sUnityAdsListener);
-        if (UnityAds.isReady()) {
+        if (UnityAds.isReady(sPlacementId)) {
```
rsatrasala: Can you use hasVideoAvailable()?

What if this method fails? Shouldn't we fail the ad request?

rsatrasala: What if this method fails? Shouldn't we fail the ad request?

chauduyphanvu1: Nope, this is an early check "just in case" there's something loaded already. And if so, simply dispatch the success callback. It's not meant to play the video from here if one has already been loaded. The normal lifecycle will actually be after this point.

chauduyphanvu1: Sure can!

PR #29, opened 2018-05-15
Make adapters GDPR-compliant

chauduyphanvu1:
Also pushed the new Vungle adapters in this PR.


AppLovin/com/mopub/mobileads/AppLovinRewardedVideo.java
```diff
@@ -48,6 +51,15 @@
...
     protected boolean checkAndInitializeSdk(@NonNull final Activity activity, @NonNull final Map<String, Object> localExtras, @NonNull final Map<String, String> serverExtras) throws Exception {
+
+        PersonalInfoManager personalInfoManager = MoPub.getPersonalInformationManager();
+
+        // Pass the user consent from the MoPub SDK to AppLovin as per GDPR
+        if (personalInfoManager != null) {
+            boolean canCollectPersonalInfo = personalInfoManager.canCollectPersonalInformation();
+            AppLovinPrivacySettings.setHasUserConsent(canCollectPersonalInfo, activity.getApplicationContext());
+        }
+
```
thomasmso: Could we bump the internal versioning of AppLovin's 3 adapters from `MoPub-Certified-32.2.2` to `MoPub-Certified-3.0.0`? Reason for the major version bump is this makes it easier for us to remember which version is GDPR-compliant.

AppLovin/com/mopub/mobileads/AppLovinBanner.java
```diff
@@ -41,6 +44,15 @@
 
     @Override
     protected void loadBanner(final Context context, final CustomEventBannerListener customEventBannerListener, final Map<String, Object> localExtras, final Map<String, String> serverExtras) {
+
+        PersonalInfoManager personalInfoManager = MoPub.getPersonalInformationManager();
+
+        // Pass the user consent from the MoPub SDK to AppLovin as per GDPR
+        if (personalInfoManager != null && personalInfoManager.gdprApplies()) {
```
bichenwang: you don't need to check if gdpr applies or not. you don't have to check if personal info manager is null. we do that all for you if you do `MoPub.canCollectPersonalInformation`

rsatrasala: @bichenwang the check for gdpr applies is necessary since there can be at least one request in un-consented state if publisher does not initialize.

rsatrasala:
Please update change-log with information related to consent. 
After that LGTM.

AppLovin/com/mopub/mobileads/AppLovinBanner.java
bichenwang: if that is true, then gdprApplies() is null, and this will crash.
Please only use MoPub.canCollectPersonalInformation unless you need the consent status. Then you can use that. You should let publishers use gdprApplies() if and only if they really just need that check.

rsatrasala:
@ratm Thanks for checking. We are working with Flurry to finalize the adapter changes. Updates will be added shortly.

OnebyAOL/com/mopub/mobileads/MillennialBanner.java
```diff
@@ -58,6 +63,20 @@ protected void loadBanner(final Context context, final CustomEventBannerListener
         if (context instanceof Activity) {
             try {
                 MMSDK.initialize((Activity) context, ActivityListenerManager.LifecycleState.RESUMED);
+
+                PersonalInfoManager personalInfoManager = MoPub.getPersonalInformationManager();
+
+                if (personalInfoManager != null) {
+                    boolean gdprApplies = personalInfoManager.gdprApplies();
```
bichenwang: this can cause a null pointer exception. gdprApplies() can return null

linked67:
Any update about Flurry ?
Is there a list of adapter That are ready for GDPR ?

PR #32, opened 2018-05-21
Pass the "npa" field from publisher's app to AdMob via ad request

chauduyphanvu1:


AdMob/com/mopub/mobileads/GooglePlayServicesInterstitial.java
```diff
@@ -67,14 +71,25 @@ protected void onInvalidate() {
         }
     }
 
+    private Bundle getConsent(Map<String, Object> localExtras) {
```
rsatrasala: Rename this method to getGooglePersonalizationPreferences. 

Add a comment that this setting is applicable to only publishers who use their own consent mechanism. Consent collected from MoPub's default consent dialogue should not used to used to pass the value of npa here

AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java
```diff
@@ -167,7 +171,7 @@ public void run() {
                             .onRewardedVideoLoadSuccess(GooglePlayServicesRewardedVideo.class, mAdUnitId);
                 } else {
                     mRewardedVideoAd
-                            .loadAd(mAdUnitId, new AdRequest.Builder().setRequestAgent("MoPub").build());
+                            .loadAd(mAdUnitId, new AdRequest.Builder().addNetworkExtrasBundle(AdMobAdapter.class, getConsent(localExtrasMap)).setRequestAgent("MoPub").build());
```
rsatrasala: check is localExtrasMap is empty

chauduyphanvu1: It's checked below in the TextUtils.isEmpty() line.

chauduyphanvu1: Also within the Android SDK, `localExtras` is passed around either as a `@NonNull` or populated with a few other key-value pairs, so it is null-safe. The only thing that might be null/empty is the `npa`, and it's being guarded in this PR.

AdMob/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 15.0.0.4
+    * Pass ad personalization preference to AdMob as per GDPR. Unlike with the other mediation adapters, publishers must specify their preference ("npa") in their app implementation. The adapter will read from a MoPub's `localExtras` and forward that onto AdMob via an AdMob ad request.
```
rsatrasala: Simplify - Append user's ad personalization preference from`localExtras` to AdMob's ad request.

AdMob/com/mopub/mobileads/GooglePlayServicesBanner.java
```diff
@@ -62,6 +65,10 @@ protected void loadBanner(
 
         final AdRequest adRequest = new AdRequest.Builder()
                 .setRequestAgent("MoPub")
+                // This setting is applicable to only publishers who use their own consent mechanism.
```
rsatrasala: Consent collected from the MoPub’s consent dialogue  should not be used to set up Google's personalization preference. Publishers should work with Google to be GDPR-compliant.

rsatrasala:
LGTM with the comment changes

PR #33, opened 2018-05-21
Pass consent to AdColony in both interstitial and rewarded video adapters

brittmullen:
AdColonyAppOptions are sent via configure with GDPR specific keys using implied explicit consent and the value from MoPub.canCollectPersonalInformation().

In the case where the adapter is initialized after AdColony is already configured, we send this same information to AdColony using the setAppOptions() method.

Edit: This PR has been updated based on feedback on the iOS side of things. We no longer will pass GDPR/consent values to AdColony when GDPR does not apply based on the personal info manager's gdprApplies() method.


rsatrasala:
Changelog updates are needed.
Otherwise LGTM

rsatrasala:
LGTM

PR #34, opened 2018-05-22
Updated GDPR and user consent setting on Tapjoy Adapter

clarachenhub:
The Tapjoy Adapter can now update the MoPub GDPR setting to Tapjoy through fetchMoPubGDPRSettings() on network events.

How to Test
Setup
- Clone Mopub Sample app from [Mopub repo](https://github.com/mopub/mopub-android-sdk).
- Follow both Mopub-Tapjoy Adapter and Tapjoy SDK Integration setup.
-  Run Mopub Sample app and observe the GDPR values. 
     - GDPR values on mopub and Tapjoy network calls should be synchronized.


Tapjoy/com/mopub/mobileads/TapjoyInterstitial.java
```diff
@@ -99,6 +101,27 @@ private void createPlacement(Context context, String placementName) {
...
+
+            if (gdprApplies) {
+                Tapjoy.subjectToGDPR(true);
+
+                String userConsent = MoPub.canCollectPersonalInformation() ? "1" : "0";
+                Tapjoy.setUserConsent(userConsent);
+
+            } else {
+                Tapjoy.subjectToGDPR(false);
+                Tapjoy.setUserConsent("-1");
```
chauduyphanvu1: @clarachenhub On this line, `gdprApplies` is going to be null (instead of `false`), but the adapter is still passing `-1`. Should there be no `else` clause here? You can check for `if gdprApplies` is `true` or `false` in the `if` clause and set `subjectToGDPR()` accordingly.

clarachenhub: Thank you very much for pointing it out. I have updated with checking gdprApplies only "true" or "false".

rsatrasala:
LGTM

PR #38, opened 2018-05-25
Guard against potential NPEs when AdMob MediationSettings are not set

chauduyphanvu1:
* Deconstruct the AdMob ad request's builder to only forward the `npa` if a Bundle is passed in. No longer passing an empty Bundle.
* Add `setNpaBundle()` for apps to call and send in a new `npa` Bundle.
* Guard against potential NPEs.


thomasmso:
@chauduyphanvu1 - Did you see my follow-up issue? https://github.com/mopub/mopub-ios-mediation/pull/61/files

PR #39, opened 2018-05-26
Fix AdMob's isLoaded() not called on main thread for mediation on Unity

chauduyphanvu1:
* This is a fix for ADF-3407.
* Replace a call to Google's `RewardedVideoAd.isLoaded()` with a local boolean set to true when MoPub's `onRewardedVideoLoadSuccess()` fires. `hasVideoAvailable` then returns that boolean.

### DIFF
--- a/AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java
+++ b/AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java
@@ -58,6 +58,11 @@ public class GooglePlayServicesRewardedVideo extends CustomEventRewardedVideo im
     private RewardedVideoAd mRewardedVideoAd;
 
     /**
+     * Flag to indicate whether the rewarded video has cached.
+     */
+    private boolean isAdLoaded;
+
+    /**
      * A {@link LifecycleListener} used to forward the activity lifecycle events from MoPub SDK to
      * Google Mobile Ads SDK.
      */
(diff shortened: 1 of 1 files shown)


AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java
```diff
@@ -294,20 +302,20 @@ private MoPubErrorCode getMoPubErrorCode(int error) {
     }
 
     public static final class GooglePlayServicesMediationSettings implements MediationSettings {
-        private Bundle npaBundle;
+        private static Bundle npaBundle;
```
sdolim: This isn't a good design.  As far as I can see, the RV Manager is still the correct place to get ahold of the global mediation settings objects.

chauduyphanvu1: We can't do it via rewarded video because the Android SDK allows "modularized" imports based on your preference (https://bintray.com/mopub/mopub-android-sdk) - so some pubs will import other ad formats except rewarded video. That causes this to crash because the adapter can't find the rewarded video API.

chauduyphanvu1: https://github.com/mopub/mopub-android-mediation/commit/972bef795519f0a1764d5bcd67c5be08b1bc3dc5 was done. Just re-adding it to this specific PR.

sdolim: The iOS SDK puts the global mediation settings on the MoPub singleton, which is always present since it is in the core SDK.  I think the Android SDK needs to follow suit so you can leverage it the same way the iOS adapter code does.  

I'd be OK with this static setup if it is a temporary workaround until the SDK makes that change.  (And it should commented as such...)

sdolim:
I'll mark it approved so you can make the call on landing the change.  Just please add the comment in the code about the static being a workaround.

chauduyphanvu1:
Thanks! Going to do a direct merge with the fix and close this PR (too much stuff needs to be accounted for otherwise).

PR #41, opened 2018-06-12
ironSource adapter update

orikn:
Add onResume() onPause() based on Mopub API
send adFailed error to Mopub SDK on show fail scenarios


IronSource/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 6.7.9.1
+    * Add Activity lifecycle listeners
```
rsatrasala: Another log "Added MoPub fail ad listener if ironSource ad fails to show"

PR #44, opened 2018-06-15
Update to share consent with Unity Ads only when user provides an explicit yes/no. In all other cases, Unity Ads SDK will collect its own consent per the guidelines [here][https://unity3d.com/legal/gdpr]

ekrzanich:
…es or no


UnityAds/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 2.2.1.2
+    * Pass explicit consent to Unity Ads (with revision).
```
rsatrasala: For parity with iOS ""Update to share consent with Unity Ads only when user provides an explicit yes/no. In all other cases, Unity Ads SDK will collect its own consent per the guidelines [here][<relevant documentation link here>]""

rsatrasala:
@ekrzanich Changes look good and I have same question about DNT here as well.

UnityAds/CHANGELOG.md
ekrzanich: Update to share consent with Unity Ads only when user provides an explicit yes/no. In all other cases, Unity Ads SDK will collect its own consent per the guidelines [here][https://unity3d.com/legal/gdpr]

ekrzanich:
We don't handle LAT differently in terms of GDPR

PR #49, opened 2018-07-09
Prepare Tapjoy adapters for release - 12.0.0.0

chauduyphanvu1:
* Add `TapjoyAdvancedBidder.java` for publishers using Advanced Bidding
* Certify with Tapjoy 12.0.0


Tapjoy/com/mopub/mobileads/TapjoyAdvancedBidder.java
```diff
@@ -0,0 +1,19 @@
...
+import android.content.Context;
+
+import com.mopub.common.MoPubAdvancedBidder;
+import com.tapjoy.Tapjoy;
+
+public class TapjoyAdvancedBidder implements MoPubAdvancedBidder {
+    @Override
+    public String getToken(final Context context) {
+        String token = Tapjoy.getUserToken();
+        return (token != null && !token.isEmpty() ? token : "1");
```
bichenwang: You can use TextUtils.isEmpty([string]) to check null and empty.

chauduyphanvu1: Yep but I didn't want to have to use another import only for that. It'd make sense if it would be used more often.

PR #53, opened 2018-07-26
Add manual impression tracking to select ad formats

chauduyphanvu1:
*AdColony*
Interstitial

*AdMob*
Banner
Interstitial

*AppLovin*
Banner
Interstitial

*Chartboost*
Interstitial

*Facebook*
Banner
Interstitial

*Flurry*
Interstitial

*ironSource*
Interstitial

*One by AOL / Millennial*
-Banner- SDK does not have an equivalent callback for impression tracking / ad opened: https://github.com/mopub/mopub-android-mediation/blob/master/OnebyAOL/com/mopub/mobileads/MillennialBanner.java#L164-L270
Interstitial

*Tapjoy*
Interstitial

*Unity*
Interstitial

*Vungle*
Interstitial

### DIFF
--- a/AdColony/com/mopub/mobileads/AdColonyInterstitial.java
+++ b/AdColony/com/mopub/mobileads/AdColonyInterstitial.java
@@ -65,6 +65,8 @@ public class AdColonyInterstitial extends CustomEventInterstitial {
             return;
         }
 
+        setAutomaticImpressionAndClickTracking(false);
+
         String clientOptions = DEFAULT_CLIENT_OPTIONS;
         String appId = DEFAULT_APP_ID;
         String[] allZoneIds = DEFAULT_ALL_ZONE_IDS;
@@ -179,6 +181,7 @@ public class AdColonyInterstitial extends CustomEventInterstitial {
                         @Override
                         public void run() {
                             mCustomEventInterstitialListener.onInterstitialShown();
+                            mCustomEventInterstitialListener.onInterstitialImpression();
                         }
                     });
                 }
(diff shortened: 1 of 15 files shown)


wacamoe:
Can we confirm that these same networks are the same ones currenly using the feature on iOS? We want to ensure we aren't adding it to ones that iOS isn't also using these impression callbacks.

chauduyphanvu1:
@wacamoe - on it! 

We could also end up doing the reverse - having more iOS adapters implement manual imp tracking, since more network SDKs will likely have updated and exposed dedicated imp callbacks, too. This list for Android is rather long but there are more network SDKs there that can do it.

AdColony/com/mopub/mobileads/AdColonyInterstitial.java
```diff
@@ -179,6 +181,7 @@ public void onOpened(@NonNull com.adcolony.sdk.AdColonyInterstitial ad) {
                         @Override
                         public void run() {
                             mCustomEventInterstitialListener.onInterstitialShown();
+                            mCustomEventInterstitialListener.onInterstitialImpression();
```
bichenwang: this actually does nothing different. onInterstitialShown() does the same exact thing as onInterstitialImpression when automatic impression and click tracking is false.
So this change is exactly the same as having automatic impression true. So let's try to limit this to networks that have this on iOS.

AdMob/com/mopub/mobileads/GooglePlayServicesBanner.java
```diff
@@ -127,6 +130,12 @@ private AdSize calculateAdSize(int width, int height) {
         /*
          * Google Play Services AdListener implementation
          */
+
+        @Override
+        public void onAdImpression() {
+            mBannerListener.onBannerImpression();
```
bichenwang: would recommend a null check on all these onBannerImpression calls

PR #54, opened 2018-07-30
Adding null check

ScottishPig:
One by AOL native ads often don't come with a rating field (it's called out in their docs -- http://docs.onemobilesdk.aol.com/android-ad-sdk/native-types.html ). Null checking is necessary for the line accessing it.


chauduyphanvu1:
This looks good to me. @rsatrasala, thoughts?

chauduyphanvu1:
Incorporated in https://github.com/mopub/mopub-android-mediation/commit/acbd705e5cc2594ffc57df1775471174c7d76c67.

PR #56, opened 2018-08-07
Vungle's AutoRotation set on by default

mihailogazda:
The Vungle ad view activity would flip the device orientation because of the missing autoRotate setting. Fixed now. Not sure why it wasn't enabled by default in the first place.


chauduyphanvu1:
Hello, @mihailogazda, we reached out to Vungle and were informed there is going to be official support for this in the future. As of right now, I'm closing the PR, but feel free to watch the repo for an upcoming adapter release that contains support for rotation.

PR #62, opened 2018-08-14
Fix rewarded video NPE when returning null third-party ad ids

chauduyphanvu1:


AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java
```diff
@@ -101,7 +101,7 @@ protected LifecycleListener getLifecycleListener() {
     protected String getAdNetworkId() {
         // Google rewarded videos do not have a unique identifier for each ad; using ad unit ID as
         // an identifier for all ads.
-        return mAdUnitId;
+        return mAdUnitId == null ? "" : mAdUnitId;
```
bichenwang: so this is great, but for the callbacks (like on line 159 for example), we're using the raw mAdUnitId which may be null and which would still cause a NPE. For this particular custom event, you can use getAdNetworkId() to be safe. Other networks can use similar things, or you can just check for null on network ids.

bichenwang:
You can mark each member variable as @NonNull, and the compiler will give
you warnings if it can be null. Your solution is perfectly valid. Add these
annotations to make sure it's truly not null.

chauduyphanvu1:
Closing due to conflicts with master. Will open a new PR.

PR #69, opened 2018-08-21
AdMob - Add manual impression tracking

chauduyphanvu1:


AdMob/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 15.0.0.10
```
rsatrasala: How about RV??
Changelog updates same as Facebook

chauduyphanvu1: Our Android API doesn't have one for rewarded video (https://github.com/mopub/mopub-android/pull/1309#issuecomment-407596113) because it doesn't make much sense to.

PR #72, opened 2018-08-22
One by AOL - Add manual impression tracking

chauduyphanvu1:


OnebyAOL/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 6.8.1.4
```
rsatrasala: Changelog updates same as Facebook

rsatrasala: Why are other format changes not included?

chauduyphanvu1: Reason was given on `EPSO-13969` but in short, the AOL SDK does not have a corresponding banner callback for an "ad show" event to map ours to (https://github.com/mopub/mopub-android-mediation/blob/master/OnebyAOL/com/mopub/mobileads/MillennialBanner.java#L141-L206).

rsatrasala: ah got it.

PR #73, opened 2018-08-23
Support flurry 11.2.0

hearsilent:


hunterhays:
Flurry 11.2.0 and adapters will work with MoPub, but only in the case that the Flurry session is started in the app's onCreate method. This is not always desirable for mediation users. For this reason at Flurry's request MoPub has have held off on certifying 11.2.0 officially.

PR #79, opened 2018-09-12
Expose the Flurry advertiser name asset

chauduyphanvu1:


Flurry/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 10.1.0.2
+    * Expose the Flurry advertiser name asset.
```
rsatrasala: Facilitate publishers to use Flurry's advertiser asset in their native ad layout

PR #81, opened 2018-09-17
Use ConsentStatus object to determine explicit consent rather than only relying on gdprApplies() and canCollectPersonalInformation()

brittmullen:


rsatrasala:
@brittmullen  we have reviewed these changes and we will be sharing the updates with the selected publishers. We are working on additional checks on our end before we can make it public

PR #83, opened 2018-09-26
Fail the Unity Ads interstitial and rewarded video adapters on no fills

fritzhuie:
When NO_FILL is detected for a given placement, the adaptor will now alert Mopub immediately.


UnityAds/com/mopub/mobileads/UnityInterstitial.java
```diff
@@ -33,6 +33,9 @@ protected void loadInterstitial(Context context,
             if (UnityAds.isReady(mPlacementId)) {
                 mCustomEventInterstitialListener.onInterstitialLoaded();
                 loadRequested = false;
+            } else if (UnityAds.getPlacementState(mPlacementId) == UnityAds.PlacementState.NO_FILL){
```
rsatrasala: Is there any other reason for timeout beyond no fill? Could other reasons for failure be possible? @fritzhuie

chauduyphanvu1: @fritzhuie ping. Could you help address @rsatrasala's question above?

fritzhuie: Hi @rsatrasala,

I'm not aware of any other scenarios that lead to a time out at the moment, but I'm happy to address other known issues as well.

No_fill scenarios are not being actively communicated to the mopub SDK through the adaptor on Android, so I think it's important to correct that.

This change should particularly improve performance for publishers using eCPM limiters to force a lower fill rate.

rsatrasala: Thank you @fritzhuie for the explanation. We will push and continue to monitor the improvements.

PR #88, opened 2018-10-09
Roll back AppLovin banner manual impression tracking

chauduyphanvu1:


AppLovin/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog
+  * 8.1.4.1
+    * Roll back the change introduced in v8.1.0.2 that aligned MoPub's banner and interstitial impression tracking to that of AppLovin.
```
rsatrasala: Just add one additional liner that - the decision was taken to avoid incorrectly counting higher impressions as `adDisplayed` callbacks are not de-duped.

PR #92, opened 2018-10-18
Adds AudienceNetworkAds.initialize call to each of the ad formats

xilosada:


FacebookAudienceNetwork/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## Changelog
+
+  * 5.0.1.0-beta
+    * Added calls to Facebook SDK initialization method for each ad format.
```
chauduyphanvu1: I'd rephrase it to `Initialize the Facebook Audience Network SDK [...]`

FacebookAudienceNetwork/com/mopub/mobileads/FacebookRewardedVideo.java
```diff
@@ -63,6 +66,9 @@ protected boolean checkAndInitializeSdk(@NonNull Activity launcherActivity, @Non
 
     @Override
     protected void loadWithSdkInitialized(@NonNull Activity activity, @NonNull Map<String, Object> localExtras, @NonNull Map<String, String> serverExtras) throws Exception {
+        if(!sIsInitialized.getAndSet(true)) {
+            AudienceNetworkAds.initialize(activity);
```
chauduyphanvu1: `loadWithSdkInitialized ` assumes that mediated SDKs have been initialized. Is this call async? If so, I'd call it in MoPub's `checkAndInitializeSdk()`.

xilosada: @chauduyphanvu1 The call is not async, but it seems that `checkAndInitializeSdk()` is the right place to do it. I moved the call to this method. Is it ok?

FacebookAudienceNetwork/CHANGELOG.md
xilosada: Updated

PR #100, opened 2018-11-06
Adds Banner support for UnityAds

RossRothenstine:
## Description

This PR will add UnityAds Banner support and does some small cleanup to the codebase. Note that UnityAds Banner currently only supports a single banner instance.

## Testing

 - Pixel 2 Android 8.0
 - Nexus 5 Tablet Android 7.0


UnityAds/com/mopub/mobileads/UnityBanner.java
```diff
@@ -0,0 +1,131 @@
...
+		this.customEventBannerListener = customEventBannerListener;
+		this.context = context;
+		UnityRouter.getBannerRouter().setCurrentPlacementId(placementId);
+
+		if (UnityRouter.initUnityAds(serverExtras, (Activity)context)) {
+			UnityRouter.getBannerRouter().addListener(placementId, this);
+			// Bug: banner ready events go through the interstitial router atm.
+			UnityRouter.getInterstitialRouter().addListener(placementId, this);
+
+			if (UnityAds.isReady(placementId)) {
```
dravyan: 1) Can you let us know if you are able to successfully load the ad in your testing?. IsReady() always returning false and hence the load banner is not being called. It looks like isInitialized() never succeeds which in turn return false.
2) Is there any specific adunit  that we can use to test it ?. We have created and using the adunit "6b89949c52cd4fa4ac50423d63123b4d". But we are failed to load the ad due to isReady() check

Can you please verify  and update us ?

RossRothenstine: Hello Dravya,

1. I was seeing the ad unit load using the sample app.
2. I used ad unit ID "e1f1cc8425324386aa7f133db6ee602b" on Android for testing. If you are getting no fill, try setting the test mode to true.

dravyan: Hi Ross,
1. In my testing, the method isReady() never returns true in the above snippet and my understanding is that the block never gets called. Could you please elaborate on why this check if required?
2. Also, can you please provide more information on why you are adding listeners for both Interstitial and Banner here?. 
```
UnityRouter.getBannerRouter().addListener(placementId, this);
// Bug: banner ready events go through the interstitial router atm.
UnityRouter.getInterstitialRouter().addListener(placementId, this);
```

UnityAds/com/mopub/mobileads/UnityBanner.java
```diff
@@ -0,0 +1,142 @@
...
+import com.unity3d.services.banners.UnityBanners;
+
+import java.util.Map;
+
+public class UnityBanner extends CustomEventBanner implements IUnityBannerListener, IUnityAdsExtendedListener {
+
+	private Context context;
+	private String placementId = "banner";
+	private CustomEventBannerListener customEventBannerListener;
+	private View view;
```
dravyan: Can you change this variable to more descriptive something like ```private View bannerView```. This will help when you make a new ad request when bannerView == null. Let me know your thoughts.

UnityAds/com/mopub/mobileads/UnityBanner.java
```diff
@@ -0,0 +1,142 @@
...
+		} else {
+			MoPubLog.e("Failed to initialize Unity Ads");
+			if (customEventBannerListener != null) {
+				customEventBannerListener.onBannerFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+			}
+		}
+	}
+
+	private void initNoRefreshMetaData(Context context) {
+		MetaData gdprMetaData = new MetaData(context);
```
dravyan: Can you rename this variable to something that doesn't have the word gdpr in it. The reason is MetaData gdprMetaData is already used appropriately in the UnityRouter for true GDPR purposes. This MetaData here, on the other hand, does not transmit any GDPR-related data to Unity Ads.

UnityAds/com/mopub/mobileads/UnityBannerCallbackRouter.java
```diff
@@ -0,0 +1,74 @@
...
+	@Override
+	public void onUnityBannerShow(String placementId) {
+		IUnityBannerListener listener = listeners.get(placementId);
+		if (listener != null) {
+			listener.onUnityBannerShow(placementId);
+		}
+	}
+
+	@Override
+	public void onUnityBannerClick(String placementId) {
```
dravyan: This callback never fires in our testing. Could you please provide more information on what circumstances this call back is fires or doesn't fire?

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
```diff
@@ -46,19 +46,18 @@ public boolean checkAndInitializeSdk(@NonNull final Activity launcherActivity,
...
                 return false;
             }
 
-            try {
-                UnityRouter.initUnityAds(serverExtras, launcherActivity);
-                UnityRouter.addListener(sPlacementId, sUnityAdsListener);
-            } catch (UnityRouter.UnityAdsException e) {
-                MoPubLog.e("Failed to initialize Unity Ads.");
-                MoPubRewardedVideoManager.onRewardedVideoLoadFailure(UnityRewardedVideo.class, sPlacementId, UnityRouter.UnityAdsUtils.getMoPubErrorCode(e.getErrorCode()));
+            UnityRouter.getInterstitialRouter().setCurrentPlacementId(sPlacementId);
```
dravyan: Looks like setCurrentPlacementId() is called from all adapters, so you can place it in the centralized UnityRouter.java` instead? Let me know your thoughts on this.

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
```diff
@@ -70,12 +69,13 @@ protected void loadWithSdkInitialized(@NonNull Activity activity,
         sPlacementId = UnityRouter.placementIdForServerExtras(serverExtras, sPlacementId);
         mLauncherActivity = activity;
 
-        UnityRouter.addListener(sPlacementId, sUnityAdsListener);
+        UnityRouter.getInterstitialRouter().addListener(sPlacementId, sUnityAdsListener);
+        UnityRouter.getInterstitialRouter().setCurrentPlacementId(sPlacementId);
```
dravyan: This is called earlier in the same class already, not sure why it's here again. Can you provide more information on this?

UnityAds/com/mopub/mobileads/UnityBanner.java
RossRothenstine: 1. For banners and interstitial, the placement is marked ready via the `onUnityAdsReady` call, unfortunately this call is always tied to interstitials, thus we subscribe to both.

2. There isn't a guarantee that `UnityAds.isReady(placementId)` will be true at this time due to how our SDK is initialized. If it's already initialized, chances are good that it will be ready here. If not, we load our banner in the `onUnityAdsReady` callback.

UnityAds/com/mopub/mobileads/UnityBanner.java
RossRothenstine: :+1: Will do.

UnityAds/com/mopub/mobileads/UnityBanner.java
RossRothenstine: :+1: Will do.

UnityAds/com/mopub/mobileads/UnityBannerCallbackRouter.java
RossRothenstine: This callback should fire when the banner is tapped. Let me double check that it's fired from the WebView.

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
RossRothenstine: We might need some clarity on the difference between `checkAndInitializeSdk` and `loadWithSdkInitialized`. It seems like we duplicate our init logic.

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
RossRothenstine: Both Banner and Interstitial have their notion of "currentPlacementId", which can both be different. I.e., a banner is loaded, then an interstitial is played. These would both be different placement IDs.

UnityAds/com/mopub/mobileads/UnityRewardedVideo.java
dravyan: checkAndInitializeSdk() -  
  ```   * Sets up the 3rd party ads SDK if it needs configuration. Extenders should use this
     * to do any static initialization the first time this method is run by any class instance.
     * From then on, the SDK should be reused without initialization.
     *
     * @return true if the SDK performed initialization, false if the SDK was already initialized.
     ```
loadWithSdkInitialized() - 
```  * Runs the ad-loading logic for the 3rd party SDK. localExtras & serverExtras should together
     * contain all the data needed to load an ad.
     * <p/>
     * Implementers should also use this method (or checkAndInitializeSdk)
     * to register a listener for their SDK.
     * <p/>
     * This method should not call any {@link MoPubRewardedVideoManager} event methods directly
     * (onAdLoadSuccess, etc). Instead the SDK delegate/listener should call these methods.```

So, you just need to register listeners in only one of the above methods. That should be fine.

File: UnityAds/com/mopub/mobileads/UnityRouter.java [hunk 7]
```diff
@@ -1,64 +1,65 @@
...
+    }
+
+    static void initGdpr(Context context) {
         // Pass the user consent from the MoPub SDK to Unity Ads as per GDPR
         PersonalInfoManager personalInfoManager = MoPub.getPersonalInformationManager();
 
         if (personalInfoManager != null) {
             ConsentStatus consentStatus = personalInfoManager.getPersonalInfoConsentStatus();
 
             if(consentStatus == ConsentStatus.EXPLICIT_YES || consentStatus == ConsentStatus.EXPLICIT_NO) {
```
rsatrasala: Change this logic to use `canCollectPersonalInfo`

dravyan: Ok updated PR to use canCollectPersonalInfo

PR #98, opened 2018-11-05
Add null check before calling isLoaded() for RewardedVideos for Admob

dravyan:

PR #105, opened 2018-11-15
Check for null before mRewardedVideoAd.loadAd

joeykrim:
Fix a received NPE crash at com.mopub.mobileads.GooglePlayServicesRewardedVideo$2.run (GooglePlayServicesRewardedVideo:190), which is mRewardedVideoAd.loadAd(mAdUnitId, adRequest), by confirming mRewardedVideoAd is not null before calling.

Similar to the null check added a few days ago on first part of the if statement: https://github.com/mopub/mopub-android-mediation/pull/98


chauduyphanvu1:
Thanks for the PR! The issue might be deeper due to multithreading, so I'm going to move everything right above the main looper `Runnable` to within it (instead of keeping adding null checks, since the rewarded ad will be guaranteed to be non null).

joeykrim:
Should be addressed better by new PR: https://github.com/mopub/mopub-android-mediation/pull/121

PR #110, opened 2018-12-05
[Android] Ready to release IronSource adapters 6.8.0.1.0

dravyan:


IronSource/CHANGELOG.md
```diff
@@ -1,7 +1,4 @@
 ## Changelog
-  * 6.8.0.1.0
```
chauduyphanvu1: This needs to be put back because the changelog from master went in first, which contained this.

PR #119, opened 2018-12-27
Fix NPE calling RewardedVideoAd loadAd() method

evabishchevich:
Looks like https://github.com/mopub/mopub-android-mediation/pull/98 is not enough. We still try to load ad on null object


chauduyphanvu1:
Thanks for the PR! The issue might be deeper due to multithreading, so I'm going to move everything right above the main looper `Runnable` to within it (instead of returning if the rewarded ad is null, we will have the chance to create it anew and use it going forward).

PR #120, opened 2019-01-01
Null check to avoid NPEs for mRewardedVideoAd.loadAd calls

joeykrim:
When using separate threads, avoid NPEs on calls to mRewardedVideoAd.loadAd(mAdUnitId, adRequest) (https://github.com/mopub/mopub-android-mediation/blob/master/AdMob/com/mopub/mobileads/GooglePlayServicesRewardedVideo.java#L197)


joeykrim:
image

PR #148, opened 2019-03-12
Remove the /libs dir since we are no longer distributing jars

chauduyphanvu1:
* Remove the '/libs/ dir that holds jars for Unity build scripts
* Once we are certain no other build script uses that dir, we can merge this PR


aornelas:
let's hold off on this untill we discuss the broader implications in the next working group!

chauduyphanvu1:
We are using `libs/` to hold released AARs, so it's still in use. Closing the PR.

PR #152, opened 2019-03-27
Add Verizon adapters for external code review

chauduyphanvu1:
* Retain the `OnebyAOL` directory and retrofit select Millennial adapter files to subclass their Verizon counterparts
* Add Verizon adapter source files to a `Verizon` directory
* Add changelog
* Format code
* Remove bid code and return `null` for the bidding token
* Pass MoPub's log level to Verizon
* Improve initialization experience in the AdapterConfiguration
* Add null checks for listeners
* Remove duplicated callback


Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+import static com.mopub.common.logging.MoPubLog.AdapterLogEvent.CUSTOM_WITH_THROWABLE;
+
+public class VerizonAdapterConfiguration extends BaseAdapterConfiguration {
+
+    public static final String ADAPTER_VERSION = BuildConfig.VERSION_NAME;
+    public static final String MEDIATOR_ID = "MoPubVAS-" + ADAPTER_VERSION;
+
+    private static final String ADAPTER_NAME = VerizonAdapterConfiguration.class.getSimpleName();
+    private static final String MOPUB_NETWORK_NAME = BuildConfig.NETWORK_NAME;
+    private static final String SITE_ID = "siteId";
```
rjmtweep: I suggest refactoring the name to `SITE_ID_KEY`

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                        MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+
+                if (bannerListener != null) {
+                    bannerListener.onBannerFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+                }
+
+                return;
+            }
+
+            final boolean success = StandardEdition.initializeWithActivity((Activity) context, siteId);
```
rjmtweep: Why are we only handling initialization with an `Activity` here and never an `Application`? 

Why do we use `StandardEdition` to initialize here, but use `VASAds` to initialize in the adapter config?

Verizon/src/main/java/com/mopub/mobileads/VerizonUtils.java
```diff
@@ -0,0 +1,30 @@
...
+        handler.post(runnable);
+    }
+
+    static MoPubErrorCode convertErrorCodeToMoPub(final int vasErrorCode) {
+        switch (vasErrorCode) {
+            case ERROR_NO_FILL:
+                return MoPubErrorCode.NETWORK_NO_FILL;
+            case ERROR_AD_REQUEST_TIMED_OUT:
+                return MoPubErrorCode.NETWORK_TIMEOUT;
+            case ERROR_AD_REQUEST_FAILED:
```
bichenwang: Maybe also MoPubErrorCode. NETWORK_NO_FILL? Unspecified is maybe okay in this case, but it feels more like a no fill. That or NETWORK_INVALID_STATE?

settings.gradle
```diff
@@ -11,7 +11,8 @@ include ':AdColony',
         ':OnebyAOL',
         ':Tapjoy',
         ':UnityAds',
-        ':Vungle'
+        ':Vungle',
```
bichenwang: verizon should go above Vungle

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                        listener.onBannerFailed(errorCode);
+                    }
+                    MoPubLog.log(SHOW_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
+                }
+            });
+        }
+
+        @Override
+        public void onResized(final InlineAdView inlineAdView) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Verizon banner resized");
```
bichenwang: should we print out what size it got resized to?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+
+    @Override
+    protected void loadBanner(final Context context, final CustomEventBannerListener customEventBannerListener,
+                              final Map<String, Object> localExtras, final Map<String, String> serverExtras) {
+
+        bannerListener = customEventBannerListener;
+
+        if (serverExtras == null || serverExtras.isEmpty()) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Ad request to Verizon failed because " +
+                    "serverExtras is null");
```
bichenwang: is null or empty

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
bichenwang: what's the difference between standard edition initialize and VASAds init?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                return;
+            }
+
+            internalView = new FrameLayout(context);
+
+            LayoutParams lp = new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT);
+            lp.gravity = Gravity.CENTER_HORIZONTAL;
+            internalView.setLayoutParams(lp);
+
+            AdViewController.setShouldHonorServerDimensions(internalView);
```
bichenwang: I don't think this view has been measured or laid out. Are you sure you want to pass this view to setShouldHonorServerDimension? Or do the only things that matter are the flags for match_parent and wrap_content and the gravity?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+            inlineAdFactory.setRequestMetaData(requestMetadata);
+            inlineAdFactory.load(new VerizonInlineAdListener());
+
+            verizonAdapterConfiguration.setCachedInitializationParameters(context, serverExtras);
+        } catch (NumberFormatException e) {
+            MoPubLog.log(CUSTOM_WITH_THROWABLE, "Fail to parse ad's width and/or height.", e);
+
+            if (bannerListener != null) {
+                bannerListener.onBannerFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+            }
```
rjmtweep: This is the seventh place in this method were we have the same 3 lines of code. We should create a local "fail" method that does the null check and calls `onBannerFailed` with `ADAPTER_CONFIGURATION_ERROR` and/or check as many of these values at possible earlier in the method (using something like `extrasAreValid`) and remove several of these redundant lines.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+    @Override
+    protected void loadBanner(final Context context, final CustomEventBannerListener customEventBannerListener,
+                              final Map<String, Object> localExtras, final Map<String, String> serverExtras) {
+
+        bannerListener = customEventBannerListener;
+
+        if (serverExtras == null || serverExtras.isEmpty()) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Ad request to Verizon failed because " +
+                    "serverExtras is null");
+            MoPubLog.log(LOAD_FAILED, ADAPTER_NAME, MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR.getIntCode(),
```
bichenwang: Do keep in mind we only have int codes for network_timeout, adapter_not_found, and ad_success. Everything else will just be 10000.

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+
+        if (networkInitializationSucceeded) {
+            listener.onNetworkInitializationFinished(VerizonAdapterConfiguration.class,
+                    MoPubErrorCode.ADAPTER_INITIALIZATION_SUCCESS);
+        } else {
+            listener.onNetworkInitializationFinished(VerizonAdapterConfiguration.class,
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+        }
+
+        final MoPubLog.LogLevel mopubLogLevel = MoPubLog.getLogLevel();
```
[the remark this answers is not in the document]
rjmtweep: Agreed. Especially since there are so many logs in the Verizon Ads initialization method.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+
+            final InlineAdFactory inlineAdFactory = new InlineAdFactory(context, placementId,
+                    Collections.singletonList(new AdSize(width, height)), new VerizonInlineAdFactoryListener());
+
+            final RequestMetadata requestMetadata = new RequestMetadata.Builder().setMediator(VerizonAdapterConfiguration.MEDIATOR_ID).build();
+            inlineAdFactory.setRequestMetaData(requestMetadata);
+            inlineAdFactory.load(new VerizonInlineAdListener());
+
+            verizonAdapterConfiguration.setCachedInitializationParameters(context, serverExtras);
+        } catch (NumberFormatException e) {
```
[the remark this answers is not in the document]
rjmtweep: Agreed. There's no fallback logic to handle the case where we get a `NumberFormatException`, so there's no reason to put it anywhere other than right after the parsing calls.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+
+                    if (listener != null) {
+                        listener.onBannerLoaded(internalView);
+                    }
+                }
+            });
+        }
+
+        @Override
+        public void onCacheLoaded(final InlineAdFactory inlineAdFactory, final int i, final int i1) {
```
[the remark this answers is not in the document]
rjmtweep: What are these for?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        @Override
+        public void onCacheUpdated(final InlineAdFactory inlineAdFactory, final int i) {
```
rjmtweep: What is `i`?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+            VerizonUtils.postOnUiThread(new Runnable() {
+
+                @Override
+                public void run() {
+                    final MoPubErrorCode errorCode = VerizonUtils.convertErrorCodeToMoPub(errorInfo.getErrorCode());
+
+                    if (listener != null) {
+                        listener.onBannerFailed(errorCode);
+                    }
+                    MoPubLog.log(SHOW_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
```
rjmtweep: This should be before the `onBannerFailed` callback.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
```
rjmtweep: This should be before the `onBannerFailed` callback.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
rjmtweep: This is in the `VerizonInlineAdFactoryListener`. Shouldn't this be `LOAD_FAILED`?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+
+        @Override
+        public void onAdLeftApplication(final InlineAdView inlineAdView) {
+            // Only logging this event. No need to call bannerListener.onLeaveApplication()
+            // because it's an alias for bannerListener.onBannerClicked()
+            MoPubLog.log(WILL_LEAVE_APPLICATION, ADAPTER_NAME);
+        }
+
+        @Override
+        public void onAdRefreshed(final InlineAdView inlineAdView) {
```
rjmtweep: When is this called? Do we need to log anything here?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        @Override
+        public void onEvent(final InlineAdView inlineAdView, final String s, final String s1, final Map<String, Object> map) {
```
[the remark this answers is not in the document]
rjmtweep: As well as a name for the map that gives some indication as to the contents.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+    @NonNull
+    private VerizonAdapterConfiguration verizonAdapterConfiguration;
+
+    public VerizonInterstitial() {
+        verizonAdapterConfiguration = new VerizonAdapterConfiguration();
+    }
+
+    @Override
+    protected void loadInterstitial(final Context context,
+                                    final CustomEventInterstitialListener customEventInterstitialListener, final Map<String, Object> localExtras,
```
rjmtweep: Split these onto two lines.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+        final String placementId = serverExtras.get(PLACEMENT_ID_KEY);
+
+        if (TextUtils.isEmpty(placementId)) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Ad request to Verizon failed because placement " +
+                    "ID is empty");
+            MoPubLog.log(LOAD_FAILED, ADAPTER_NAME, MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR.getIntCode(),
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+
+            if (interstitialListener != null) {
+                interstitialListener.onInterstitialFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
```
rjmtweep: Just as in the banner class, there are so many instances of these same 3 lines. Refer to those comments for ways to address this.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+        interstitialAdFactory.setRequestMetaData(requestMetadata);
+        interstitialAdFactory.load(new VerizonInterstitialListener());
+
+        verizonAdapterConfiguration.setCachedInitializationParameters(context, serverExtras);
+    }
+
+    @Override
+    protected void showInterstitial() {
+
+        MoPubLog.log(SHOW_ATTEMPTED, ADAPTER_NAME, "Showing Verizon interstitial");
```
rjmtweep: `SHOW_ATTEMPTED` should only be used with the `ADAPTER_NAME`. No other strings should be passed.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+            public void run() {
+                if (verizonInterstitialAd != null) {
+                    verizonInterstitialAd.show(context);
+                    return;
+                }
+
+                if (interstitialListener != null) {
+                    interstitialListener.onInterstitialFailed(MoPubErrorCode.INTERNAL_ERROR);
+                }
+                MoPubLog.log(SHOW_FAILED, ADAPTER_NAME, MoPubErrorCode.UNSPECIFIED, "Failed to " +
```
rjmtweep: Needs to be before the `onInterstitialFailed` callback.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+            VerizonUtils.postOnUiThread(new Runnable() {
+
+                @Override
+                public void run() {
+                    final MoPubErrorCode errorCode = VerizonUtils.convertErrorCodeToMoPub(errorInfo.getErrorCode());
+
+                    if (listener != null) {
+                        listener.onInterstitialFailed(errorCode);
+                    }
+                    MoPubLog.log(LOAD_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
```
rjmtweep: Needs to be before `onInterstitialFailed`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+            VerizonUtils.postOnUiThread(new Runnable() {
+
+                @Override
+                public void run() {
+                    final MoPubErrorCode errorCode = VerizonUtils.convertErrorCodeToMoPub(errorInfo.getErrorCode());
+
+                    if (listener != null) {
+                        listener.onInterstitialFailed(errorCode);
+                    }
+                    MoPubLog.log(SHOW_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
```
rjmtweep: Needs to be before the callback.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+                public void run() {
+                    if (listener != null) {
+                        listener.onInterstitialLoaded();
+                    }
+                }
+            });
+        }
+
+        @Override
+        public void onCacheLoaded(final InterstitialAdFactory interstitialAdFactory, final int i, final int i1) {
```
rjmtweep: Parameter names aren't descriptive enough.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        @Override
+        public void onCacheUpdated(final InterstitialAdFactory interstitialAdFactory, final int i) {
```
rjmtweep: Parameter names aren't descriptive enough.

Verizon/src/main/java/com/mopub/mobileads/VerizonUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import android.os.Looper;
+
+public final class VerizonUtils {
+
+    private static final Handler handler = new Handler(Looper.getMainLooper());
+
+    // VASAds error codes
+    private static final int ERROR_NO_FILL = -1;
+    private static final int ERROR_AD_REQUEST_TIMED_OUT = -2;
+    private static final int ERROR_AD_REQUEST_FAILED = -3;
```
rjmtweep: Remove these and add these imports:

```
import static com.verizon.ads.VASAds.ERROR_AD_REQUEST_FAILED;
import static com.verizon.ads.VASAds.ERROR_AD_REQUEST_TIMED_OUT;
import static com.verizon.ads.VASAds.ERROR_NO_FILL;
```

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
```diff
@@ -0,0 +1,27 @@
+package com.mopub.mobileads;
+
+import com.verizon.ads.Bid;
+import com.verizon.ads.Configuration;
+import com.verizon.ads.support.TimedMemoryCache;
+
+final class BidCache {
+
+    private static final int SUPER_AUCTION_CACHE_TIMEOUT_DEFAULT = 600000;
```
rjmtweep: Needs a comment specifying the unit of measurement. Is this 600k milliseconds or microseconds (or something else)?

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String DOMAIN = "com.verizon.ads";
+    private static final String SUPER_AUCTION_CACHE_TIMEOUT_KEY = "super.auction.cache.timeout";
```
rjmtweep: These timeout keys are private. Do we really need such verbose names? Shouldn't `TIMEOUT_KEY` and `TIMEOUT_DEFAULT` work?

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private static TimedMemoryCache<Bid> bidTimedMemoryCache;
+
+    static {
+        bidTimedMemoryCache = new TimedMemoryCache<>();
+    }
+
+    static void put(final String placementId, final Bid bid) {
+        bidTimedMemoryCache.add(placementId, bid,
+                (long) Configuration.getInt(DOMAIN, SUPER_AUCTION_CACHE_TIMEOUT_KEY, SUPER_AUCTION_CACHE_TIMEOUT_DEFAULT));
```
rjmtweep: Long names make this messy. It might be worth breaking it up onto multiple lines just to make it cleaner (or reduce the length of the names).

Ex:

```
final long timelimit = (long) Configuration.getInt(DOMAIN, 
            SUPER_AUCTION_CACHE_TIMEOUT_KEY, 
            SUPER_AUCTION_CACHE_TIMEOUT_DEFAULT));
bidTimedMemoryCache.add(placementId, bid, timelimit);
```

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -158,12 +160,18 @@ protected void loadBanner(final Context context, final CustomEventBannerListener
             AdViewController.setShouldHonorServerDimensions(internalView);
             VASAds.setLocationEnabled(MoPub.getLocationAwareness() != MoPub.LocationAwareness.DISABLED);
 
+            Bid bid = BidCache.get(placementId);
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -175,6 +183,40 @@ protected void loadBanner(final Context context, final CustomEventBannerListener
...
+    public static void requestBid(final Context context, final String placementId, final List<AdSize> adSizes,
+                                  final RequestMetadata requestMetadata, final BidRequestListener bidRequestListener) {
+
+        if (bidRequestListener == null) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "bidRequestListener parameter cannot be null.");
+
+            return;
+        }
+
+        RequestMetadata.Builder builder = new RequestMetadata.Builder(requestMetadata);
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -175,6 +183,40 @@ protected void loadBanner(final Context context, final CustomEventBannerListener
...
+        RequestMetadata actualRequestMetadata = builder.setMediator(VerizonAdapterConfiguration.MEDIATOR_ID).build();
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -130,13 +132,53 @@ protected void loadInterstitial(final Context context,
         final InterstitialAdFactory interstitialAdFactory = new InterstitialAdFactory(context, placementId,
                 new VerizonInterstitialFactoryListener());
 
-        final RequestMetadata requestMetadata = new RequestMetadata.Builder().setMediator(VerizonAdapterConfiguration.MEDIATOR_ID).build();
-        interstitialAdFactory.setRequestMetaData(requestMetadata);
-        interstitialAdFactory.load(new VerizonInterstitialListener());
+        Bid bid = BidCache.get(placementId);
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -130,13 +132,53 @@ protected void loadInterstitial(final Context context,
...
+    public static void requestBid(final Context context, final String placementId, final RequestMetadata requestMetadata,
+                                  final BidRequestListener bidRequestListener) {
+
+        if (bidRequestListener == null) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "bidRequestListener parameter cannot be null.");
+
+            return;
+        }
+
+        RequestMetadata.Builder builder = new RequestMetadata.Builder(requestMetadata);
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -130,13 +132,53 @@ protected void loadInterstitial(final Context context,
...
+        RequestMetadata actualRequestMetadata = builder.setMediator(VerizonAdapterConfiguration.MEDIATOR_ID).build();
```
rjmtweep: `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -130,13 +132,53 @@ protected void loadInterstitial(final Context context,
...
+
+        InterstitialAdFactory.requestBid(context, placementId, actualRequestMetadata, new BidRequestListener() {
+            @Override
+            public void onComplete(Bid bid, ErrorInfo errorInfo) {
```
rjmtweep: `final` `final`

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -175,6 +183,40 @@ protected void loadBanner(final Context context, final CustomEventBannerListener
...
+
+        InlineAdFactory.requestBid(context, placementId, adSizes, actualRequestMetadata, new BidRequestListener() {
+            @Override
+            public void onComplete(Bid bid, ErrorInfo errorInfo) {
```
rjmtweep: `final` `final`

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
thomasmso: I've found that using something like `TimeUnit.MINUTES.toMillis(10)` is clear and self-documenting.

rjmtweep: I agree that it's clean and self-documenting, but `5 * 60 * 1000; // 5 minutes` would be clear enough and wouldn't require an extra import.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
chauduyphanvu1: @mmullin, could you help us with the proper API to use to initialize Verizon? I can ensure consistency afterwards.

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+                    networkInitializationSucceeded = true;
+                } else if (configuration != null && context instanceof Application) {
+                    final String siteId = configuration.get(SITE_ID);
+
+                    if (TextUtils.isEmpty(siteId)) {
+                        MoPubLog.log(CUSTOM, ADAPTER_NAME, "Verizon's initialization not " +
+                                "started. Ensure Verizon's siteId is populated on the MoPub " +
+                                "dashboard. Note that initialization on the first app launch is a no-op.");
+                    } else {
+                        VASAds.initialize((Application) context, siteId);
```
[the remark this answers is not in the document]
chauduyphanvu1: @mmullin, please let us know Verizon's preference.

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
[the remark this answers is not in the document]
chauduyphanvu1: @bichenwang, there's no observable effect doing it before or after the callback.

@mmullin, correct me if I'm wrong but Verizon does not have a "none"/"off" log level, and that the default appears to be "info". The adapter can flip through the different levels like it's doing here.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+import static com.mopub.common.logging.MoPubLog.AdapterLogEvent.LOAD_SUCCESS;
+import static com.mopub.common.logging.MoPubLog.AdapterLogEvent.SHOW_FAILED;
+import static com.mopub.common.logging.MoPubLog.AdapterLogEvent.WILL_LEAVE_APPLICATION;
+
+public class VerizonBanner extends CustomEventBanner {
+
+    private static final String ADAPTER_NAME = VerizonBanner.class.getSimpleName();
+
+    private static final String PLACEMENT_ID_KEY = "placementId";
+    private static final String SITE_ID_KEY = "siteId";
```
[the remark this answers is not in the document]
chauduyphanvu1: Going to leave this as-is because Verizon retrofitted the existing Millennial custom events to subclass the Verizon counterparts, so it might be more convenient leaving these keys in for readability purposes.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                        MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+
+                if (bannerListener != null) {
+                    bannerListener.onBannerFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+                }
+
+                return;
+            }
+
+            if (!(context instanceof Activity)) {
```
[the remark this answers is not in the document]
chauduyphanvu1: @mmullin, please let us know Verizon's preference.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                // Destroy any hanging references
+                if (verizonInlineAd != null) {
+                    verizonInlineAd.destroy();
+                    verizonInlineAd = null;
+                }
+            }
+        });
+    }
+
+    class VerizonInlineAdFactoryListener implements InlineAdFactory.InlineAdFactoryListener {
```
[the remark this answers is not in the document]
chauduyphanvu1: Going to leave it as non-static.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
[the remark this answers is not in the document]
chauduyphanvu1: @mmullin, please make the change.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,335 @@
...
+                    if (listener != null) {
+                        listener.onBannerFailed(errorCode);
+                    }
+                    MoPubLog.log(SHOW_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
+                }
+            });
+        }
+    }
+
+    private class VerizonInlineAdListener implements InlineAdView.InlineAdListener {
```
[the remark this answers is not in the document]
chauduyphanvu1: Going to leave it as non-static.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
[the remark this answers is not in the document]
chauduyphanvu1: @mmullin, please make the change.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+                // Destroy any hanging references
+                if (verizonInterstitialAd != null) {
+                    verizonInterstitialAd.destroy();
+                    verizonInterstitialAd = null;
+                }
+            }
+        });
+    }
+
+    class VerizonInterstitialFactoryListener implements InterstitialAdFactory.InterstitialAdFactoryListener {
```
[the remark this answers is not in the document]
chauduyphanvu1: Going to leave it as non-static.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
chauduyphanvu1: FYI @mmullin.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
chauduyphanvu1: FYI @mmullin.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
chauduyphanvu1: FYI @mmullin.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,311 @@
...
+                    if (listener != null) {
+                        listener.onInterstitialFailed(errorCode);
+                    }
+                    MoPubLog.log(LOAD_FAILED, ADAPTER_NAME, errorCode.getIntCode(), errorCode);
+                }
+            });
+        }
+    }
+
+    private class VerizonInterstitialListener implements InterstitialAd.InterstitialAdListener {
```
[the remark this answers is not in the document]
chauduyphanvu1: Going to leave it as non-static.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
chauduyphanvu1: FYI @mmullin.

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
chauduyphanvu1: FYI @mmullin.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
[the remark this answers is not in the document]
rjmtweep: Verizon provides methods for initialization with both `Application` context and `Activity` context; we should check what we have and use the appropriate method.

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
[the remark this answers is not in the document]
rjmtweep: Verizon has several log levels, it just happens to default to `INFO`:

```
package com.verizon.ads;

import android.util.Log;

public final class Logger {
    public static final int VERBOSE = 2;
    public static final int DEBUG = 3;
    public static final int INFO = 4;
    public static final int WARN = 5;
    public static final int ERROR = 6;
    private static final String TAG_PREFIX = "VAS-";
    private static volatile int logLevel = 4;
    private final String tag;
//...
```

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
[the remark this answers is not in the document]
rjmtweep: As I said in another comment, we should support both since Verizon supplies initialization methods for both.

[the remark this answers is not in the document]
chauduyphanvu1: The `Context` is going to always be an `Application` in this method anyway so there's no point using their `Activity` initialization method.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
mmullin: @chauduyphanvu1 @rjmtweep The Verizon Ads SDK is componentized and gives us the ability to release specific sets of plugins to different publishers based on their needs. We call these supported sets of audience specific plugins, Editions.

Standard Edition  includes support for Inline, Interstitial, and Native placements as well as support for display, video, and native ads.

The Edition itself is a plugin, and provides convenience methods for publishers to initialize their SDK. The `StandardEdition` initialization wraps the `VASAds` initialize call, and performs other tasks required for the SDK to startup properly. Using `VASAds` directly to initialize the SDK will result in a non-functioning Edition, and shouldn't be used in the adapter.

In order to allow publishers to gracefully migrate over from the Millennial SDK, the adapters should defer pre-initialization altogether for the time being and only use the `StandardEdition.initializeWithActivity(Activity activity, String siteId)` call to initialize the Standard Edition at ad request time.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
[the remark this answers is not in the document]
mmullin: The cache listeners have empty implementations because our cache mechanism is not used in the adapters at all. See our [class docs](https://sdk.verizonmedia.com/android-class-docs.html) for more details, but the 2nd and 3rd variable names for onCacheLoaded are numRequested and numReceived.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
mmullin: cacheSize

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
rsatrasala: @mmullin I am jumping in because we are also discussing about pre-initialization in the emails. Understand the significance of using `StandardEdition.initializeWithActivity(Activity activity, String siteId)`. Could we use the same in the adapterConfig also for pre-initialization ? Do you see any other limitations after reading my latest email response on how MoPub pre-init works ?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
mmullin: The onAdRefreshed listener is only called for InlineAdViews where a refresh interval has been set. Since publishers will be using the MoPub refresh mechanism, it will never be called during the mediation lifecycle

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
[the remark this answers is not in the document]
mmullin: full method -- `onEvent (InlineAdView inlineAdView, String source, String eventId, Map<String, Object> arguments)`

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
mmullin: first int is numRequested, second int is numReceived

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
mmullin: variable name is cacheSize

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
mmullin: @rsatrasala There is a limitation on our side with tracking the Activity lifecycle if initializeWithActivity is called during pre-init. That limitation also extends to our `initialize(Application application, String siteId)` call when we are a mediated SDK in the request flow. Our ask is to defer pre-initialization in the short term, and we will follow up with an update in a few weeks that adds full pre-initialization support.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
mmullin: Good catch @bichenwang ... seems like this is a remnant from our old adapter code. We've tested quickly without it, and the adapter works fine. The line can be removed

OnebyAOL/src/main/java/com/mopub/mobileads/MillennialAdapterConfiguration.java
```diff
@@ -1,85 +1,6 @@
...
-                    MoPubErrorCode.ADAPTER_INITIALIZATION_SUCCESS);
-        } else {
-            listener.onNetworkInitializationFinished(MillennialAdapterConfiguration.class,
-                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
-        }
-    }
+@SuppressWarnings("unused")
+public class MillennialAdapterConfiguration extends VerizonAdapterConfiguration {
+    // Intentionally left empty to just provide the Millennial namespaced version for backward compatibility.
 }
```
rjmtweep: Missing newline at end of file.

OnebyAOL/src/main/java/com/mopub/mobileads/MillennialBanner.java
```diff
@@ -1,285 +1,35 @@
...
-        public void onResized(InlineAd inlineAd, int w, int h, boolean isClosed) {
-            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Millennial Inline Ad - Banner resized (width: " + w + ", " +
-                    "height: " + h + "). " + (isClosed ? "Returned to original placement." :
-                    "Got a fresh, new place."));
-        }
+    @Override
+    protected String getHeightKey() {
+        return "adHeight";
     }
 }
```
rjmtweep: Missing newline at end of file.

OnebyAOL/src/main/java/com/mopub/mobileads/MillennialBanner.java
```diff
@@ -1,285 +1,35 @@
...
-    private static final String AD_HEIGHT_KEY = "adHeight";
-
-    private InlineAd inlineAd;
-    private CustomEventBannerListener bannerListener;
-    private FrameLayout internalView;
-
-    static {
-        MoPubLog.log(CUSTOM, ADAPTER_NAME, "Millennial Media Adapter Version: " + MillennialUtils.MEDIATOR_ID);
+    public MillennialBanner() {
+        MoPubLog.log(CUSTOM, ADAPTER_NAME, "Verizon Adapter Version: " + VerizonAdapterConfiguration.MEDIATOR_ID);
```
rjmtweep: Why do we have a log here? The Verizon adapter has no log and the INIT_FINISHED log includes Verizon adapter info.

OnebyAOL/src/main/java/com/mopub/mobileads/MillennialBanner.java
```diff
@@ -1,285 +1,35 @@
...
 
-    private CreativeInfo getCreativeInfo() {
-        if (inlineAd == null) {
-            return null;
-        }
-
-        return inlineAd.getCreativeInfo();
+    @Override
+    protected String getPlacementIdKey() {
+        return "adUnitID";
```
rjmtweep: 1. Let's keep this consistent with the Verizon and keep these in `private static final String` fields instead of putting the literal in the method.

OnebyAOL/src/main/java/com/mopub/mobileads/MillennialInterstitial.java
```diff
@@ -1,367 +1,25 @@
...
-                        interstitialListener.onInterstitialShown();
-                        interstitialListener.onInterstitialImpression();
-                    }
-                }
-            });
-        }
+    protected String getSiteIdKey() {
+        return "dcn";
     }
 }
```
rjmtweep: Missing newline at end of file.

rjmtweep: And same comments as `MillennialBanner.java`

Verizon/README.md
```diff
@@ -0,0 +1,7 @@
+## Overview
+  * This folder contains mediation adapters used to mediate Verizon.
+  * To download and integrate the Verizon SDK, please check [this tutorial](https://sdk.verizonmedia.com/integration.html).
+  * For inquiries and support, please utilize [the support portal](https://learn.onemobile.aol.com/hc/en-us/requests/new).
+
```
rjmtweep: Should we have a comment about AB not being implemented yet, but some of the boilerplate and stub-code being present?

Verizon/src/main/java/com/mopub/mobileads/BidCache.java
rjmtweep: They're `private static final` - renaming them here will not affect any other code.

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+
+        final String adapterVersion = getAdapterVersion();
+        return (!TextUtils.isEmpty(adapterVersion)) ? adapterVersion.substring(0, adapterVersion.lastIndexOf('.')) : "";
+    }
+
+    @Override
+    public void initializeNetwork(@NonNull final Context context, @Nullable final Map<String, String> configuration,
+                                  @NonNull final OnNetworkInitializationFinishedListener listener) {
+
+        Preconditions.checkNotNull(listener);
```
rjmtweep: Missing `Preconditions.checkNotNull(context);`

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,403 @@
...
+
+        bannerListener = customEventBannerListener;
+
+        if (serverExtras == null || serverExtras.isEmpty()) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Ad request to Verizon failed because " +
+                    "serverExtras is null or empty");
+
+            logAndNotifyBannerFailed(LOAD_FAILED,
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR.getIntCode(),
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
```
rjmtweep: This is so much better! But it can be cleaned up even more:

```
import static com.mopub.mobileads.MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR;
// any other error code static imports needed

// ...

logAndNotifyBannerFailed(LOAD_FAILED, MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);

// ...

    private void logAndNotifyBannerFailed(final MoPubLog.AdapterLogEvent event, final MoPubErrorCode errorCode) {

        MoPubLog.log(event, ADAPTER_NAME, errorCode.getIntCode(), errorCode);

        if (bannerListener != null) {
            bannerListener.onBannerFailed(errorCode);
        }
    }
```

Note: The `finals` in the method signature should be added even if you don't make the rest of the changes here.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,403 @@
...
+            return;
+        }
+
+        try {
+            adWidth = Integer.parseInt(widthString);
+            adHeight = Integer.parseInt(heightString);
+        } catch (NumberFormatException e) {
+            MoPubLog.log(CUSTOM_WITH_THROWABLE, "Failed to parse ad's width and/or height.", e);
+
+            logAndNotifyBannerFailed(LOAD_FAILED, MoPubErrorCode.INTERNAL_ERROR.getIntCode(),
```
rjmtweep: There should be a `return` here, right?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,403 @@
...
+                    MoPubErrorCode.INTERNAL_ERROR);
+        }
+
+        if (adWidth < 0 || adHeight < 0) {
```
rjmtweep: Only `< 0`? What if one of them *is* 0? Isn't that still a failure?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -0,0 +1,403 @@
...
+                    logAndNotifyBannerFailed(SHOW_FAILED, errorCode.getIntCode(), errorCode);
+                }
+            });
+        }
+
+        @Override
+        public void onResized(final InlineAdView inlineAdView) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Verizon banner resized to: " +
+                    inlineAdView.getAdSize().getWidth() + " by " +
+                    inlineAdView.getAdSize().getHeight());
```
rjmtweep: `getAdSize()` can return null - need to add a check.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,350 @@
...
+
+        interstitialListener = customEventInterstitialListener;
+        this.context = context;
+
+        if (serverExtras == null || serverExtras.isEmpty()) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Ad request to Verizon failed because " +
+                    "serverExtras is null or empty");
+
+            logAndNotifyInterstitialFailed(LOAD_FAILED, MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR.getIntCode(),
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
```
rjmtweep: The refactor suggestions from **VerizonBanner.java** apply here too.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,350 @@
...
+     *
+     * @param context            a non-null Context
+     * @param placementId        a valid placement ID. Cannot be null or empty.
+     * @param requestMetadata    a {@link RequestMetadata} instance for the request or null
+     * @param bidRequestListener an instance of {@link BidRequestListener}. Cannot be null.
+     */
+    public static void requestBid(final Context context, final String placementId, final RequestMetadata requestMetadata,
+                                  final BidRequestListener bidRequestListener) {
+
+        if (bidRequestListener == null) {
```
rjmtweep: Why are we only checking `bidRequestListener == null` when the JavaDoc specifies that `context` and `placementId` also need to be non-null (and `placementId` can't be empty)?

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -0,0 +1,350 @@
...
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "bidRequestListener parameter cannot be null.");
+
+            return;
+        }
```
rjmtweep: Why are we only checking bidRequestListener == null when the JavaDoc specifies that context and placementId also need to be non-null (and placementId can't be empty)?

settings.gradle
```diff
@@ -20,7 +21,7 @@ if (file('Testing').exists()) {
 
 // MoPub SDK
 def mopubSdkDir = file('mopub-android-sdk').exists() &&
-                  file(".gitmodules").text.contains('[submodule "mopub-android-sdk"]') ?
+        file(".gitmodules").text.contains('[submodule "mopub-android-sdk"]') ?
         'mopub-android-sdk' : 'mopub-android-sdk'
```
rjmtweep: Still not sure what purpose this serves...

chauduyphanvu1: @dravyan, does removing this affect any other thing you are working on? Is there a benefit to have this on the public repo?

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
chauduyphanvu1: We're not using the `context` anymore since we removed the initialization code here. I'll add it back if we will be doing initialization.

Verizon/README.md
chauduyphanvu1: Sounds good to note about that. @rsatrasala, what do you say?

settings.gradle
dravyan: @chauduyphanvu1, pushed fix in the latest commit. Can you revert this change here?

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -193,11 +186,15 @@ public static void requestBid(final Context context, final String placementId,
...
-        }
+        // Fail fast here on null. The Verizon Ads SDK checks for empty downstream.
+        Preconditions.checkNotNull(context, "Super auction bid skipped because the " +
+                "context is null");
+        Preconditions.checkNotNull(placementId, "Super auction bid skipped because the " +
+                "placement ID is null");
+        Preconditions.checkNotNull(adSizes, "Super auction bid skipped because the " +
+                "adSizes list is null");
+        Preconditions.checkNotNull(bidRequestListener, "Super auction bid skipped " +
+                "because the bidRequestListener is null");
```
rjmtweep: I thought we were avoiding the use of `Preconditions` in adapters themselves.  Additionally, `placementId` and `adSizes` need to be checked for emptiness.

Verizon/src/main/java/com/mopub/mobileads/VerizonInterstitial.java
```diff
@@ -143,11 +142,13 @@ protected void loadInterstitial(final Context context,
...
-
-            return;
-        }
+        // Fail fast here on null. The Verizon Ads SDK checks for empty downstream.
+        Preconditions.checkNotNull(context, "Super auction bid skipped because context " +
+                "is null");
+        Preconditions.checkNotNull(placementId, "Super auction bid skipped because the " +
+                "placement ID is null");
+        Preconditions.checkNotNull(bidRequestListener, "Super auction bid skipped because " +
+                "the bidRequestListener is null");
```
rjmtweep: `placementId` needs to be checked for emptiness.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
chauduyphanvu1: The Verizon Ads SDK checks for emptiness later on so I did not want to complicate this part of the method just to do elaborate validation controls. I didn't hear about `Preconditions` usage in adapters - we have [a few adapters](https://github.com/mopub/mopub-android-mediation/search?utf8=%E2%9C%93&q=preconditions&type=) that use it.

rjmtweep: That's my mistake, I think. I could be thinking of the decision to remove uses of Preconditions from the Sample App.

As for the emptiness, the JavaDoc specifies that it cannot be empty. We should make the behavior consistent with the documentation by either removing that from the doc or by adding a check.

PR #156, opened 2019-04-10
AdMob adapter throws exception on OnResume and OnPause. Vungle autorotation enabled.

mihailogazda:


chauduyphanvu1:
@mihailogazda, thank you for the PR! We are fixing the AdMob `onResume()` and `onPause()` issue by [removing them](https://github.com/mopub/mopub-unity-mediation/issues/34#issuecomment-479126799) since they are no longer needed in the new AdMob rewarded video API.

Regarding the Vungle rotation change, could you elaborate on the context behind that change? 

Vu Chau
MoPub SDK Team

mihailogazda:
Hey @chauduyphanvu1 Vungle one is an oldie, but from what i can see they haven't fixed it. More info here: https://github.com/mopub/mopub-android-mediation/pull/56

Vungle/src/main/java/com/mopub/mobileads/VungleInterstitial.java
```diff
@@ -117,6 +117,9 @@ public void run() {
             Object ordinalViewCount = localExtras.get(ORDINAL_VIEW_COUNT_KEY);
             if (ordinalViewCount instanceof Integer)
                 mAdConfig.setOrdinal((Integer) ordinalViewCount);
+			
+			//	Set autorotation
+			mAdConfig.setAutoRotate(true);
```
chauduyphanvu1: @ashinagawa, could you please help confirm this change? Is it a recommended fix from Vungle? If not, what is an alternative?

palrahul: @chauduyphanvu1  we do not want this to be always set to true for adapter and instead take the auto-rotate as another local extra and then pass that to our SDK. We plan to submit a PR from Vungle by early next week.
@mihailogazda thanks for bringing this to our notice

chauduyphanvu1:
Thanks for the review, @palrahul. Closing this PR in favor of the upcoming one.

mihailogazda:
Cheers guys, I'll keep an eye for the updates.

PR #165, opened 2019-05-21
Adding support for legacy MillennialMedia Custom Event inline integrations

nimrich:


Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -105,19 +107,37 @@ protected void loadBanner(final Context context,
...
+            }
         }
 
-        if (localExtras.get(getHeightKey()) != null) {
-            adHeight = (int) localExtras.get(getHeightKey());
+        if (adHeight <= 0 || adWidth <= 0) {
+            // Fallback to server extras for legacy custom event integrations
+            final String widthString = serverExtras.get(WIDTH_LEGACY_KEY);
+            final String heightString = serverExtras.get(HEIGHT_LEGACY_KEY);
+            try {
```
chauduyphanvu1: Add new line above.

Verizon/src/main/java/com/mopub/mobileads/VerizonBanner.java
```diff
@@ -105,19 +107,37 @@ protected void loadBanner(final Context context,
         }
 
         if (localExtras == null || localExtras.isEmpty()) {
-            MoPubLog.log(CUSTOM, ADAPTER_NAME, "localExtras is null. Unable to extract banner sizes");
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "localExtras is null. Unable to extract banner sizes from localExtras");
```
chauduyphanvu1: I'd recommend making the log statement so that the adapter will now try with `serverExtras`. Otherwise, this looks to be the end of the ad request logic if someone is to read logs.

nimrich:
@chauduyphanvu1 addressed your feedback.  Thanks!

Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -48,10 +48,11 @@ public String getMoPubNetworkName() {
     @NonNull
     @Override
     public String getNetworkSdkVersion() {
-        final SDKInfo sdkInfo = VASAds.getSDKInfo();
 
-        if (sdkInfo != null) {
-            return sdkInfo.version;
+        final String editionVersion = Configuration.getString("com.verizon.ads", "editionName", null);
```
chauduyphanvu1: @nimrich, I'm getting `null` for `editionVersion`. Do you know if that's intended? I'm getting `1.1.0` from `VASAds.getSDKInfo().version`, which is also incorrect (it should be `1.1.1`).

nimrich: @chauduyphanvu1 I did make a correction where this pulls "editionVersion" rather than "editionName."  That being said, it is still null prior to the SDK being initialized.  This is expected.  Once initialized, it returns "1.1.1" as expected.  VASAds.getSDKInfo().version returns only the version of our "core" module (which is 1.1.0).  We will be addressing this in a subsequent release so that VASAds.getSDKInfo().version returns the meaningful version.

PR #181, opened 2019-07-10
Release Vungle adapter 6.4.11.0

palrahul:
Adapter changes to support upcoming version of Vungle Android SDK Version 6.4.11

Change log for SDK version 6.4.11:
https://github.com/Vungle/Android-SDK/blob/6.4.11/CHANGELOG.md


Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -0,0 +1,299 @@
...
+            MoPubLog.log(mPlacementId, LOAD_ATTEMPTED, ADAPTER_NAME);
+        }
+    }
+
+    private AdConfig.AdSize getVungleAdSize(Map<String, Object> localExtras) {
+        AdConfig.AdSize adSizeType = AdConfig.AdSize.VUNGLE_DEFAULT;
+        int adWidthInDp = localExtras.containsKey(KEY_AD_WIDTH) ? (int)localExtras.get(KEY_AD_WIDTH) : 0;
+        int adHeightInDp = localExtras.containsKey(KEY_AD_HEIGHT) ? (int)localExtras.get(KEY_AD_HEIGHT) : 0;
+
+        if(adWidthInDp >= 300 && adHeightInDp >= 250) {
```
dravyan: Since you are supporting only MREC, the only change here is to include `>=`.

palrahul: @dravyan Does MoPub do any additional validation on its side when Publishers passes or sets a View container of specific size on Dashboard or Integration (getting View container sizes). As this specific check would work iff Publisher has these LocalExtras configured, was wondering if the Publishers has set specific width and height that is out of range ( say much lesser than 300,250) will Vungle adapter be even invoked in the waterfall or does MoPub filter it out based on our current supported AdSize ( currently only MREC, Banners to follow in later release)
If Vungle's adapter will get invoked in the Ad size mismatch scenario, is there any metrics exposed/public that would suggest how many of these no-fills are attributed to this mismatch for sizes that we currently dont support.
We do send onBannerFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR) but nothing that suggests AdSize mismatch

hmertkahraman: Hi @palrahul,
Chiming in to let you know of our new plan for this specific check.
We just initiated the work to convert this from a range check to parse a response coming from adserver that indicates the creative type, example of the json would be:
 `adunit-format`: `medium_rectangle`
and so we will get a property such as ` adUnitFormat` from `local_extras` and see if it matches `medium_rectangle` to handle this check.
When we're able to get that, we won't need to check for range scenarios. If the creative type doesn't match the existing adapter, which is currently mrect only, we will simply fail there.

On AdSize mismatch, we do not have an error and I will pass this as a feedback.

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -0,0 +1,299 @@
...
+        public void onAdEnd(@NonNull String placementReferenceId, boolean wasSuccessfulView, final boolean wasCallToActionClicked) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "onAdEnd placement id"+ placementReferenceId);
+            if (mPlacementId.equals(placementReferenceId)) {
+                MoPubLog.log(CUSTOM, ADAPTER_NAME, "onAdEnd - Placement ID: " + placementReferenceId + ", wasSuccessfulView: " + wasSuccessfulView + ", wasCallToActionClicked: " + wasCallToActionClicked);
+                mIsPlaying = false;
+
+                mHandler.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        if (wasCallToActionClicked) {
```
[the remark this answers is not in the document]
palrahul: @hmertkahraman 
Reg 
> Can there be any edge cases affecting preventing this from healthily working (not counting a click event accurately)? 
None that we have come across in our testing, onAdEnd will get called consistently when the Ad is finished, in refresh scenario as well as when the parent view container / controller is dismissed. We use this same value in our internal reporting so don't expect any discrepancy.

Reg
> On other networks clicks are usually counted as soon as the click action occurs, a callback informs us to fire the click trackers right away.
This is not our current implementation, as this requires changes in our core SDK we are considering this in our next SDK release iteration , and corresponding adapter releases. We understand the need for firing events corresponding to Click in realtime, and will keep you updated on this.

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
rsatrasala: @palrahul Updating Mert's previous comment as our plan has slightly changed for Android. 
We will be updating the original check you added to include 336*280 as well.

There will be two sizes we are supporting for medium rectangles and adding both of them to ensure you get all medium rectangle traffic. Please check the updated code and let us know if you have any questions

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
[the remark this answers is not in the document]
rsatrasala: @palrahul Considering onAdEnd is not the same click callback, we recommend removing the click tracking altogether. Click callback is optional and so we can add the click callback again when you are able to track it. Please let us know if that works for you.

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -100,7 +100,22 @@ public void run() {
             mVungleAdapterConfiguration.setCachedInitializationParameters(context, serverExtras);
         }
 
-        adConfig.setAdSize(getVungleAdSize(localExtras));
+        AdConfig.AdSize vungleAdSize = getVungleAdSize(localExtras);
```
palrahul: thanks for this refactor 👍

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -118,14 +133,13 @@ public void run() {
     }
 
     private AdConfig.AdSize getVungleAdSize(Map<String, Object> localExtras) {
-        AdConfig.AdSize adSizeType = AdConfig.AdSize.VUNGLE_DEFAULT;
         int adWidthInDp = localExtras.containsKey(KEY_AD_WIDTH) ? (int)localExtras.get(KEY_AD_WIDTH) : 0;
         int adHeightInDp = localExtras.containsKey(KEY_AD_HEIGHT) ? (int)localExtras.get(KEY_AD_HEIGHT) : 0;
-
-        if(adWidthInDp >= 300 && adHeightInDp >= 250) {
-            adSizeType = AdConfig.AdSize.VUNGLE_MREC;
+        if((adWidthInDp == 300 && adHeightInDp == 250) || (adWidthInDp == 336 && adHeightInDp == 280)) {
```
palrahul: * whats the context behind adding `(adWidthInDp == 336 && adHeightInDp == 280)` for size validation, is that part of standard or the max bounds for MREC that MoPub supports
* For 336,280 will our view container for MREC be aligned in center

rsatrasala: yes and yes 

We are adding a new MRECT supported size of 336*280 and so added to the list of the requests you receive and yes it will be centered.

matteinn:
Is this good to be released? Sorry for the request but I've been waiting for 1 month 😄 thank you

dravyan:
> Is this good to be released? Sorry for the request but I've been waiting for 1 month 😄 thank you
Hi @matteinn, I am in the process of final sanity check. This version of adapters should be available by EOD. Thanks.

PR #189, opened 2019-08-13
Facebook adapters release - 5.5.0.0

xilosada:
Preparing 5.5.0 release on August 19


FacebookAudienceNetwork/src/main/java/com/mopub/mobileads/FacebookAdapterConfiguration.java
```diff
@@ -66,32 +65,30 @@ public void initializeNetwork(@NonNull final Context context,
...
     }
 
     private void refreshBidderToken(final Context context) {
         if (isComputingToken.compareAndSet(false, true)) {
             new Thread(new Runnable() {
                 @Override
                 public void run() {
-                    tokenReference.set(BidderTokenProvider.getBidderToken(context));
+                    String token = BidderTokenProvider.getBidderToken(context);
+                    if(token != null) {
```
rsatrasala: @xilosada what is the reason to remove this null check? Would be good to access previously cached token if token is not generated for some reason. Thoughts?

PR #191, opened 2019-08-15
Added extended listener to rewarded and interstitial formats

xilosada:
When the launcher activity launch mode is single task, FAN formats are not calling MoPub dismiss/close methods. This PR replaces the default Interstitial and Rewarded Video Listeners with an Extended version that has a callback to dismiss the ad when the Activity is destroyed.


JanPollack:
We're also interested in getting this!

PR #196, opened 2019-09-24
AdColony - Adapter release - 4.1.0.1

akashadcolony:


hmertkahraman:
Hi @akashadcolony,

Can we please request test app / zone id combination that returns Android Banner ads, for us to test these changes?

Thanks,
Mert
cc: @mahesh-kokate

mahesh-kokate:
@hmertkahraman You can use the following IDs:

App ID: app185a7e71e1714831a49ec7
Zone ID: vz785bc8d42d9c43fdaf

AdColony/src/main/java/com/mopub/mobileads/AdColonyAdapterConfiguration.java
```diff
@@ -112,21 +132,118 @@ public void initializeNetwork(@NonNull final Context context,
...
+        if (TextUtils.isEmpty(appId)) {
+            MoPubLog.log(CUSTOM,null,"AdColony cannot configure without a valid appId parameter. " +
+                    "Please ensure you enter a valid appId.");
+            return;
+        }
+
+        AdColonyAppOptions mAdColonyAppOptions = getAdColonyAppOptionsAndSetConsent(clientOptions);
+
+        if (!isAdColonyConfigured()) {
+            configureAdColony(context, mAdColonyAppOptions, appId, allZoneIds);
```
rajeshadc: I think below this line, can add the following line "previousAdColonyAllZoneIds = allZoneIds;" so it will not go into "else if" block and call configure again for the second time otherwise previousAdColonyAllZoneIds will get null here and calls configure again.

hmertkahraman: Great one, thanks!

AdColony/src/main/java/com/mopub/mobileads/AdColonyAdapterConfiguration.java
```diff
@@ -159,18 +157,18 @@ protected static AdColonyAppOptions getAdColonyAppOptionsAndSetConsent(String cl
...
+                    adColonyAppOptions.setGDPRConsentString("1");
                 }
-            } else {
-                adColonyAppOptions.setOption(CONSENT_GIVEN, true)
-                        .setOption(CONSENT_RESPONSE, canCollectPersonalInfo);
+            } else if (canCollectPersonalInfo) {
+                adColonyAppOptions.setGDPRConsentString("1");
+            }else {
+                adColonyAppOptions.setGDPRConsentString("0");
             }
```
hmertkahraman: @rajeshadc do you think we need to add an `else` statement here and call: `adColonyAppOptions.setGDPRRequired(false);` with `false`?

hmertkahraman:
End to end testing on MoPub side is complete.

PR #197, opened 2019-09-25
Fix medium banners for facebook adapter

Gaelan-Bolger:
Medium banners case is never hit because 250 is > 90


chauduyphanvu:
@Gaelan-Bolger, thanks for the pull request! Testing this fix out now, but logic-wise it looks good!

Vu Chau
MoPub SDK Team

PR #200, opened 2019-09-30
Updates to support Verizon v1.2.0

nimrich:


nimrich:
@chauduyphanvu thanks for taking a quick look!

It looks like the Verizon placement being used needs to be changed.  This MoPub placement ID:  9ebf4105cb2b4803a49102eef84a605f should work (this points to Verizon Site ID: 8f61f7f6d1a84bd7a633ec99db00efe1 Placement ID: 240360).  

Let me know if that works or if you have any other questions, thanks!

chauduyphanvu:
@nimrich, thanks for the new ad unit ID. I confirmed that it worked.

Out of curiosity, why would the other ad unit / site ID not work with VAS 1.2.0 but work with 1.1.4? Does the 1.2.0 release have any breaking changes with regard to ad unit / site IDs and their setup?

nimrich:
@chauduyphanvu yes, there is a breaking change for the native protocol for the 1.2.0 release.  We will be migrating existing pubs to new placement IDs as they adopt v1.2.0.

PR #204, opened 2019-10-11
Adding SDK edition name to the Verizon SDK version string

nimrich:


Verizon/src/main/java/com/mopub/mobileads/VerizonAdapterConfiguration.java
```diff
@@ -51,9 +51,12 @@ public String getNetworkSdkVersion() {
...
         final String editionVersion = Configuration.getString("com.verizon.ads",
                 "editionVersion", null);
+                
+        final String editionName = Configuration.getString("com.verizon.ads",
+                "editionName", null);
 
-        if (!TextUtils.isEmpty(editionVersion)) {
-            return editionVersion;
+        if (!TextUtils.isEmpty(editionVersion) && !TextUtils.isEmpty(editionName)) {
+            return editionName + "-" + editionVersion;
```
chauduyphanvu: @nimrich, do you need this info logged for any particular reason or use case (because it's only ever consumed by the MoPub SDK as a numerical value so we'd prefer it without alpha characters)? 

Also, if the edition name is empty (how often would that be?), our SDK won't get to log the accurate version here and will fall back to dropping the last digit off of the adapter version.

nimrich: The edition name helps us identify which SDK is being used as we have plans to roll out multiple editions in the near future.

As long as the SDK has been initialized, the edition name should not be empty so the fall back logic should not be used in normal circumstances.

chauduyphanvu: Oh I see iOS is already [doing that](https://github.com/mopub/mopub-ios-mediation/blob/master/Verizon/VerizonAdapterConfiguration.m#L75).

Assuming you will be doing measurements via console log, then that's fine because I don't see any Verizon API using this data.

nimrich: @chauduyphanvu the intent was to provide a more accurate Verizon SDK version in anticipation of rolling out multiple editions.  As you said though, our SDK does not actually consume this info.  If keeping it numerical is better for the MoPub SDK, that's fine, and we can close this PR.  

As always, appreciated your help!

chauduyphanvu: We can have it alphanumerically to avoid any edition confusions in the future. I'll prep this for release. Thanks!

PR #210, opened 2019-10-28
Updates for Verizon SDK v1.2.1

nimrich:


chauduyphanvu:
@nimrich, thanks for the PR! The SDK changelog noted `Removed all permissions from internal manifest except for INTERNET`, so should we do the same for the permissions in the [adapter AndroidManifest](https://github.com/mopub/mopub-android-mediation/blob/master/Verizon/src/main/AndroidManifest.xml)?

nimrich:
@chauduyphanvu good call!  I went ahead and updated the AndroidManifest.xml.  Sorry I missed that!

PR #216, opened 2019-12-17
Updates for Verizon SDK v1.3.0

jonalter:


chauduyphanvu:
Thanks for the PR - we'll hold onto it for now (subsequent adapter certifications will bypass the code change proposed here). We don't want to prematurely remove Advanced Bidding logic from the adapters while Verizon works on an SDK-level alternative. Let us know once the latter is done, and we can then proceed with this PR.

PR #223, opened 2020-01-31
Release Vungle adapter 6.5.2.0 

palrahul:
Adapter V6.5.2.0 compatible with Vungle's SDK V6.5.2

[Updated Feb21] Point to Vungle SDK V6.5.2, Additional logic to map requested Banner size to Vungle's supported Banner sizes


Vungle/src/main/java/com/mopub/mobileads/VungleRewardedVideo.java
```diff
@@ -44,15 +41,6 @@
...
     private static final String ADAPTER_NAME = VungleRewardedVideo.class.getSimpleName();
-    public static final String USER_ID_KEY = "userId";
-    public static final String TITLE_KEY = "title";
-    public static final String BODY_KEY = "body";
-    public static final String CLOSE_BUTTON_TEXT_KEY = "closeButtonText";
-    public static final String KEEP_WATCHING_BUTTON_TEXT_KEY = "keepWatchingButtonText";
-    public static final String IS_SOUND_ENABLED_KEY = "isSoundEnabled";
-    public static final String FLEX_VIEW_CLOSE_TIME_KEY = "flexViewCloseTimeInSec";
-    public static final String ORDINAL_VIEW_COUNT_KEY = "ordinalViewCount";
-    public static final String AUTO_ROTATE_ENABLED_KEY = "autoRotateEnabled";
```
dravyan: @palrahul , are there any changes to MediationSettings? Or is this branch not on the latest master?

dravyan: I found the new addition of `VungleMediationConfiguration` and moved all this to the new class. Can you also add this update to the changelog?

palrahul: @dravyan this was a refactor that helped us consolidate some of the code for reusability, 
sure it makes sense as we are trying to deprecate VungleMediationSettngs that we should document it in changelog

Vungle/src/main/java/com/mopub/mobileads/VungleRewardedVideo.java
```diff
@@ -226,89 +201,42 @@ private boolean validateIdsInServerExtras(Map<String, String> serverExtras) {
...
+                MoPubRewardedVideoManager.getGlobalMediationSettings(VungleMediationConfiguration.class);
+        if (globalMediationSettings == null) {
+            // pubs are using elder adapter version
+            globalMediationSettings = MoPubRewardedVideoManager.getGlobalMediationSettings(VungleMediationSettings.class);
+        }
+        VungleMediationConfiguration instanceMediationSettings =
+                MoPubRewardedVideoManager.getInstanceMediationSettings(VungleMediationConfiguration.class, mAdUnitId);
+        if (instanceMediationSettings == null) {
+            instanceMediationSettings = MoPubRewardedVideoManager.getInstanceMediationSettings(VungleMediationSettings.class, mAdUnitId);
+        }
```
dravyan: this should be `VungleMediationConfiguration` instead of `VungleMediationSettings` here.

@palrahul, is there any reason to move `VungleMediationSettings` to a separate class?. It looks like the MediationSettings is moved to another class with the name `VungleMediationConfiguration` and the functionality still remains the same? Can you confirm?

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -150,11 +179,18 @@ public void run() {
...
+            adSizeType = VUNGLE_MREC;
+        } else if (adWidthInDp == BANNER_SHORT.getWidth() && adHeightInDp == BANNER_SHORT.getHeight()) {
+            adSizeType = BANNER_SHORT;
+        } else if (adWidthInDp == BANNER.getWidth() && adHeightInDp == BANNER.getHeight()) {
+            adSizeType = BANNER;
+        } else if (adWidthInDp == BANNER_LEADERBOARD.getWidth() && adHeightInDp == BANNER_LEADERBOARD.getHeight()) {
+            adSizeType = BANNER_LEADERBOARD;
         }
+
+        return adSizeType;
```
dravyan: With this above block, Vungle will only receive ad request if there publisher specificed size is exactly 320x50 , 300x250 and 728x90. If the publisher make a request using manual sizes or using one of the [supported MoPub constants ](https://github.com/mopub/mopub-android-sdk/blob/dfd5633950314041c57dbecc21e30a037075af2d/mopub-sdk/mopub-sdk-base/src/main/java/com/mopub/mobileads/MoPubView.java#L77-L87), those requests will not go through here.
Hence you can take advantage of checking the requested size and verify if Banner can fill in with one of the supported sizes. Refer to an example [here](https://github.com/mopub/mopub-android-mediation/blob/805bb8e5b78ddb556891df1693d0b8eedbba46a6/AdMob/src/main/java/com/mopub/mobileads/GooglePlayServicesBanner.java#L189-L200) 

how to request banner using MoPub API:  https://developers.mopub.com/publishers/android/banner/#step-2-load-an-ad-into-the-banner-slot

Vungle/src/main/java/com/mopub/mobileads/VungleRewardedVideo.java
palrahul: @dravyan yes we are trying to move our logic from inner class VungleRewardedVieo#VungleMediationSettings to a standalone VungleMediationConfiguration that now is used across Rewarded, Interstitial. Its a refactor for better reusability.

The code block above seems right, we are first checking if we can get global and instance settings using our new class VungleMediationConfiguration, if not then to still support old integrations and not break them we for now we use VungleMediationSettings as a fallback. Eventually as we remove the deprecated/older VungleMediationSettings we would not need the fallback code/logic

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
palrahul: @dravyan we have updated our code with the suggestion, please revisit this part

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -288,27 +350,50 @@ public void run() {
...
-                                if (vungleBannerAd != null) {
-                                    final View adView = vungleBannerAd.renderNativeView();
-                                    if (adView != null) {
-                                        isSuccess = true;
-                                        layout.addView(adView);
-                                        mCustomEventBannerListener.onBannerLoaded(layout);
-                                        MoPubLog.log(getAdNetworkId(), LOAD_SUCCESS, ADAPTER_NAME);
+
+                                //Fix for Unity Player that can't render a view with a state changed from INVISIBLE to VISIBLE.
+                                //TODO: Remove once it's fixed in MoPub Unity plugin.
```
dravyan: @palrahul , could you share more details on what needs to be fixed in MoPuh Unity plugin here?.
Is banner ads issue on Unity fixed from Vungle side? Are there any open issues here still?

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -150,11 +180,34 @@ public void run() {
...
+                adSizeType = VUNGLE_MREC;
+            }
+        } else {
+            if (adWidthInDp >= BANNER_LEADERBOARD.getWidth() && adHeightInDp >= BANNER_LEADERBOARD.getHeight()) {
+                adSizeType = BANNER_LEADERBOARD;
+            } else if (adWidthInDp >= BANNER.getWidth() && adHeightInDp >= BANNER.getHeight()) {
+                adSizeType = BANNER;
+            } else if (adWidthInDp >= BANNER_SHORT.getWidth() && adHeightInDp >= BANNER_SHORT.getHeight()) {
+                adSizeType = BANNER_SHORT;
+            }
```
dravyan: @palrahul In iOS adapter default fallback size in Vungle short banner. Here also we need tp add the same and fallback to BANNER_SHORT, if the size requested doesn;t fall into any of the buckets to maintain the parity

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
palrahul: @dravyan this is related to an issue we had for MREC where Vungle can play Video inside WebView and the View visibility transition was causing a video rendering glitch on MoPub Unity that we could overcome by this change
We have seen similar change on AdMob unity e.g. https://github.com/googleads/googleads-mobile-unity/blob/master/source/android-library/app/src/main/java/com/google/unity/ads/Banner.java#L143

PR #227, opened 2020-02-12
perform improvement & change of banner callback

Mintegral-official:
There are two parts of this pr:

1. Make sdk init only once.  In present adapter, sdk init will be executed again every time loading ads.

2. Notice: our sdk has added “onCloseBanner” method in BannerAdListener since 12.1.51 version, this may cause some compile error. Pls don’t forget to implement this callback method in banner adapter when updating to 12.1.51 version.


chauduyphanvu:
@Mintegral-official, the title mentions "change of banner callback" but I don't see any code change related to that. Is that intended?

PR #231, opened 2020-02-21
Fix NPE crash issue while requesting ad from onExpiring callback.

rajeshadc:


hmertkahraman:
Shipped on our end [here](https://github.com/mopub/mopub-android-mediation/commit/f2fcea1001a91ef73149761e16bdd610bff87e56)

PR #248, opened 2020-03-26
Mintegral - 13.0.41.0

chauduyphanvu:


Mintegral/src/main/java/com/mopub/mobileads/MintegralAdapterConfiguration.java
```diff
@@ -117,6 +121,11 @@ public static void configureMintegral(String appId, String appKey, Context conte
                 sdk.init(mtgConfigurationMap, context);
             }
 
+            final boolean canCollectPersonalInfo = MoPub.canCollectPersonalInformation();
```
Mintegral-official: These gdpr code should be put before sdk.init().

`if (sdk != null) {
            final boolean canCollectPersonalInfo = MoPub.canCollectPersonalInformation();
            final int switchState = canCollectPersonalInfo ? IS_SWITCH_ON : IS_SWITCH_OFF;
            sdk.setUserPrivateInfoType(context, AUTHORITY_ALL_INFO, switchState);

            final Map<String, String> mtgConfigurationMap = sdk.getMTGConfigurationMap(appId, appKey);

            if (context instanceof Activity) {
                sdk.init(mtgConfigurationMap, ((Activity) context).getApplication());
            } else if (context instanceof Application) {
                sdk.init(mtgConfigurationMap, context);
            }

            sdkInitialized = true;`

PR #254, opened 2020-04-28
[Chartboost] Hotfix adapters 8.0.2.1

stosiarz:
Due to changes in the Chartboost SDK 8.0+, delegate is being nullified when switching activities. This behaviour results in blocking the Mopub SDK flow which can be recovered only by restarting the application. 

To reproduce: 
Cache and show interstitial ads multiple time.
Cache and show rewarded ads multiple times. 
Notice that at some point callbacks like dismiss are not notified and user cannot cache or show anymore. 

Fix: 
Re-attach chartboost delegate before new ad cached.


hmertkahraman:
[Merged this through an internal release](https://github.com/mopub/mopub-android-mediation/commit/269bbba9a2a125fe32f7217b4d89b773f361f3a5).

PR #257, opened 2020-05-18
Release Vungle adapter 6.7.0.0

HuFangshuai:
Vungle Android MoPub Adapter 6.7.0 Release.
- This version of adapter is compatible with Vungle 6.7.0 build.
- Support Header Bidding Participation.
- Ad playback callback event changes.


Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
```diff
@@ -276,21 +270,24 @@ private String getAdNetworkId() {
...
+            if (mPlacementId.equals(placementId)) {
+                MoPubLog.log(getAdNetworkId(), CUSTOM, ADAPTER_NAME, "onAdEnd placement id: " + placementId);
                 mIsPlaying = false;
                 sVungleRouter.removeRouterListener(mPlacementId);
                 mVungleRouterListener = null;
+            }
+        }
+
+        @Override
+        public void onAdClick(String placementId) {
```
dravyan, line 283: 6.6.1 SDK will start supporting click. callback?

Vungle/src/main/java/com/mopub/mobileads/VungleRouter.java
```diff
@@ -333,4 +365,35 @@ private void onAdAvailabilityUpdate(String placementReferenceId, boolean isAdAva
             }
         }
     };
+
+    // might be called on pubs side with header bidding and pre init Vungle sdk
```
dravyan, line 369: Could you elaborate more on this?. Is this configuration passed during MoPub initialization?

Vungle/src/main/java/com/mopub/mobileads/VungleBanner.java
HuFangshuai: Yes, from 6.6.x adapter we are supporting these callbacks: onAdClick, onAdRewarded and onAdLeftApplication.
Changelog updated.

Vungle/src/main/java/com/mopub/mobileads/VungleRouter.java
HuFangshuai: I read publisher MoPub AB integration guide from https://developers.mopub.com/publishers/advanced-bidding/#step-1-initialize-your-network-sdks and update this code part changes for the purpose of:

1. Keep the original process when initializing MoPub. Publisher can integrate this new release as usual. Such as pass configuration during MoPub initialization.
2. Support initialize Vungle SDK before MoPub. This is one new recommend process following MoPub AB guideline and lets Vungle SDK able to initialized as early as possible, and also it's ok if publisher not following AB integration guideline. The configuration here will be passed before MoPub initialization.
3. Reuse this method to make sure publisher less depends on Vungle native SDK APIs and hide some details such as: disableBannerRefresh, User-Agent header.

Vungle/build.gradle
```diff
@@ -9,6 +9,6 @@ dependencies {
     if (sdkSrcDependency) {
         implementation project(':vungle-android-sdk:publisher-sdk-android')
     } else {
-        implementation "com.vungle:publisher-sdk-android:${sdkVersion}"
+        implementation "com.github.vungle:vungle-android-sdk:${sdkVersion}"
```
dravyan, line 12: Can you help to confirm which dependency we should recommend publishers to use? `implementation 'com.github.vungle:vungle-android-sdk:6.7.0'` or `implementation 'com.vungle:publisher-sdk-android:6.7.0`?

[Vungle documentation](https://support.vungle.com/hc/en-us/articles/360002922871#option-1-gradle-integration-0-5) suggests to use `com.vungle:publisher-sdk-android`

HuFangshuai: Hi @dravyan , we use `com.vungle:publisher-sdk-android:6.7.0`. I will update this line when our sdk live on bintray.

PR #258, opened 2020-05-25
[Chartboost] Update SDK to 8.0.3 and fix rewarded ad null activity issue

stosiarz:
We discovered that rewarded ads in Chartboost adapter were throwing an error.

com.chartboost.sdk.sample.cbtest I/MoPub: [com.mopub.mobileads.MoPubRewardedVideoManager][onAdSuccess] SDK Log - Could not load custom event because Activity reference was null. Call MoPub#updateActivity before requesting more rewarded ads.

To fix it I followed common pattern from other adapters to hold activity weak reference and check it before playing rewarded video.

Chartboost SDK 8.0.3 changes:

Increase WEB_VIEW_PAGE_LOAD_TIMEOUT timeout from 3 to 15 seconds to give more time for ads to load, especially on the first try. This was causing issues where sdk could be stuck and never recover its state.

Fix callback issue for Android 4. Not supported devices weren't sending any callback which could lead to increased latency for mediations.

SDK was not able to recover its state after the device suffered from no connectivity state. This caused increased latency and removed the possibility to display ads until the app was restarted.


chauduyphanvu:
Done in https://github.com/mopub/mopub-android-mediation/commit/6902c72b38f6304bf3bec72d59f3cdff8ceedcee. Thanks!

PR #282, opened 2020-08-12
Vungle - 6.7.1.0 release

anton-blashchuk-zoolatech:
Vungle 6.7.1.0 release; change to mopub-android-sdk is due to switch from v5.13.0 to v5.13.1, which should be in mopub:master anyway, since it's an important hotfix.


matteinn:
This is now available @anton-blashchuk-zoolatech

anton-blashchuk-zoolatech:
> This is now available @anton-blashchuk-zoolatech

Thank you @matteinn !
Updated PR.

anton-blashchuk-zoolatech:
@dravyan why did you remove mopub-android-sdk hash altogether instead of just reverting to previous version(5.13.0)? I believe this is not intended operation as .gitmodules was not updated and scripts were not updated too. overall adapter won't build in this configuration since dependencies are missing.

chauduyphanvu:
@anton-blashchuk-zoolatech, thanks for the notice. We'll check what we can do here.

Vu Chau
MoPub Team

PR #283, opened 2020-08-13
[Chartboost] CharboostBanner crashes on Android 4

stosiarz:
ChartboostBanner works only with Android Api >= 21. Below that we are not able to display the Banner and we found that the current version of the adapter crashes on devices with Android 4. Added check no to perform the banner load for api below lollipop.


hmertkahraman:
Hi @stosiarz , thanks for the PR. We'll queue this for the next sprint (next week).

hmertkahraman:
Released internally [here](https://github.com/mopub/mopub-android-mediation/commit/fbcde5080c7069798bbf725c329d73c32fc21968)

PR #287, opened 2020-08-26
AppLovin: Remove MREC

p158276:


AppLovin/src/main/java/com/mopub/mobileads/AppLovinBanner.java
```diff
@@ -92,6 +93,27 @@ protected void load(@NonNull final Context context, @NonNull final AdData adData
             return;
         }
 
+        String adUnitFormat = extras.get(ADUNIT_FORMAT);
+        if (!TextUtils.isEmpty(adUnitFormat)) {
+            adUnitFormat = adUnitFormat.toLowerCase();
+        }
+
+        final boolean isBannerFormat = "banner".equals(adUnitFormat);
```
rsatrasala, line 101: check if this is "Banner".equals(adUnitFormat)
cc: @chauduyphanvu

AppLovin/src/main/java/com/mopub/mobileads/AppLovinBanner.java
```diff
@@ -277,9 +299,6 @@ private AppLovinAdSize appLovinAdSizeFromAdData(@NonNull final AdData adData) {
                 // Size can contain an AppLovin leaderboard ad size of 728x90
                 if (width >= 728 && height >= 90) {
                     adSize = AppLovinAdSize.LEADER;
-                } else if (width >= 300 && height >= 250) {
-                    // Size can contain an AppLovin medium rectangle
-                    adSize = AppLovinAdSize.MREC;
```
thomasmso, line 282: Shouldn't this be enough? And if the returned `adSize` is `nil`, we'll error out the ad load?

p158276: The ALAdSize [would default set to standard banner size](https://github.com/mopub/mopub-android-mediation/blob/de9bf7c6e462e1f4c48ae704c407236cea346c73/AppLovin/src/main/java/com/mopub/mobileads/AppLovinBanner.java#L292). It would set to leader when the width >= 728 and height >= 90. Base on the logic here, the ALAdSize won't be nil.

thomasmso: Fair

PR #302, opened 2020-10-26
Release first public version of Snap adapters

rsatrasala:


Snap/build.gradle
```diff
@@ -0,0 +1,18 @@
+project.version = '1.0.2.0'
```
rsatrasala, line 1: To do
Add changelog to talk about about the first release

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+    protected boolean checkAndInitializeSdk(@NonNull Activity launcherActivity,
+                                            @NonNull AdData adData) {
+        return false;
+    }
+
+    @Override
+    protected void load(@NonNull Context context, @NonNull AdData adData) {
+        Preconditions.checkNotNull(context);
+        Preconditions.checkNotNull(adData);
+
```
hmertkahraman, line 73: Should we add `setAutomaticImpressionAndClickTracking(false);`

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR.getIntCode(),
+                    MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+            if (mLoadListener != null) {
+                mLoadListener.onAdLoadFailed(MoPubErrorCode.ADAPTER_CONFIGURATION_ERROR);
+            }
+            return;
+        }
+
+        mSlotId = extras.get(SLOT_ID_KEY);
+        if (!TextUtils.isEmpty(mSlotId)) {
```
hmertkahraman, line 87: Not a must but ideally we can wrap this with a `try` `catch` to handle if `mSlotId` is empty and error out (if Snap doesn't respond for `mSlotId` is empty or null.

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+        mSnapAdapterConfiguration.setCachedInitializationParameters(context, extras);
+        snapAdKit.loadAds();
+
+        MoPubLog.log(getAdNetworkId(), LOAD_ATTEMPTED, ADAPTER_NAME);
+    }
+
+    @Override
+    protected void show() {
+        try {
+            snapAdKit.playAds();
```
hmertkahraman, line 148: We can have the log statement execute before the ad show logic

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+                    "Audience Network Ads");
+            MoPubLog.log(getAdNetworkId(), SHOW_FAILED, ADAPTER_NAME,
+                    MoPubErrorCode.NETWORK_NO_FILL.getIntCode(),
+                    MoPubErrorCode.NETWORK_NO_FILL);
+
+            if (mInteractionListener != null) {
+                mInteractionListener.onAdFailed((NETWORK_NO_FILL));
+            }
+
+            throw exception;
```
hmertkahraman, line 161: AFAIK we're not propogating / throwing the exception on `show()` methods, instead just handling the exception and returning. We might consider if this can affect publishers in any way when `show()` is called and this error is not expected to be handled.

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    protected void onInvalidate() {
+        // no-op
+    }
+
+    @NonNull
+    @Override
+    protected String getAdNetworkId() {
+        return mSlotId;
```
hmertkahraman, line 173: This can be null. We can consider null checking here with:
```
return mPlacementId != null ? mPlacementId : "";
```
at least for our log statements

Snap/src/main/java/com/mopub/mobileads/SnapAdInterstitial.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    @Override
+    protected boolean checkAndInitializeSdk(@NonNull Activity launcherActivity,
+                                            @NonNull AdData adData) {
+        return false;
+    }
+
+    @Override
+    protected void load(@NonNull Context context, @NonNull AdData adData) {
```
hmertkahraman, line 70: What if Snap is not initialized until this point? `checkAndInitializeSdk` is returning false and has empty body. Shouldn't we check here:
`if snap not initialized, then initialize it first`?

rsatrasala (commented):
LGTM

PR #306, opened 2020-11-20
[WIP] Google - DV360

JillSong:


rsatrasala:
@JillSong Keeping this PR open and not reviewing yet until we finalize all the details.

chauduyphanvu (commented):
@JillSong, thanks for the PR. I edited it a little bit and will begin testing soon. Did you plan to also make the changes for native ad?

JillSong:
I missed the native ad. Will soon add another commit. Thanks!

chauduyphanvu:
@JillSong - I'm not actually getting a token back. Did you mean to call `run()` on the `FutureTask`?

Also, it might not be best to do an async operation here, because MoPub will not wait for the operation (we will make an ad request regardless, so the token will be empty). We should refactor that token getter to return something within 1 ms.

JillSong:
get() is equivalent of run() and returns the result that FutureTask, and was deprecated in Java 10.

We had previously communicated with MoPub how to handle this async operation, whether to force synchronicity to change from your end to have an async interface method, and the feedback was to force synchronicity. 

To reduce the time of token getter would mean to pre-generate the signals. I thought MoPub already does this by pre-calling getBiddingToken(). Caching the signals can be done in the adapter upon initialization, or in MoPub's SDK. Please let me know why MoPub has to wait for the operation then we can weigh pros and cons. Thanks!

chauduyphanvu:
@JillSong, MoPub doesn't call `getBiddingToken()` until the ad request, so if the `FutureTask` takes time to complete, then the token will not be available (because MoPub will not wait). Is it possible for Google to have a token within 1ms the first time MoPub queries for it? And you can cache that for subsequent calls.

Here's the starting point that eventually calls `getBiddingToken()`:
https://github.com/mopub/mopub-android-sdk/blob/e8b8d865535855dacd65fdf741e1c0d5b7797ca5/mopub-sdk/mopub-sdk-base/src/main/java/com/mopub/common/AdUrlGenerator.java#L271.

Are you able to start generating and caching a token somewhere in `initializeNetwork()`?

https://github.com/mopub/mopub-android-mediation/blob/5c41838fa70a431582d8922afead885d1c82a85f/AdMob/src/main/java/com/mopub/mobileads/GooglePlayServicesAdapterConfiguration.java#L70

JillSong:
Thanks for pointing the issues out. I've pushed 3 fixes:
1. Generate query at initialization time.
2. Keep a pre-cached token.
3. Fix the use of FutureTask.

chauduyphanvu:
@JillSong, thanks for the commits!

Could you take another look at the imports and type declarations? The PR currently does not compile.

JillSong:
I just compiled and tested again with the test app you sent. It works.

PR #309, opened 2020-12-03
Supports AdMob Adaptive Banner

1008k:
Supports AdMob Adaptive Banner.
Use it as follows.

```
Map<String, Object> localExtras = new HashMap<>();
localExtras.put("adaptive_banner", true);
mMoPubView.setLocalExtras(localExtras);
```

### DIFF
--- a/AdMob/src/main/java/com/mopub/mobileads/GooglePlayServicesBanner.java
+++ b/AdMob/src/main/java/com/mopub/mobileads/GooglePlayServicesBanner.java
@@ -57,6 +57,8 @@ public class GooglePlayServicesBanner extends BaseAd {
     private String mAdUnitId;
     private Integer adWidth;
     private Integer adHeight;
+    
+    private Boolean adaptiveBanner = false;
 
     @Override
     protected void load(@NonNull final Context context, @NonNull final AdData adData) {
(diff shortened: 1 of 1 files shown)


chauduyphanvu:
@1008k, thanks for the PR. AdMob adaptive banners _are_ supported as-is in the current adapters (no additional work is needed). Whatever size you define on the MoPub side will be passed to Google, who will return flexible banners that can fill the available slot. We don't need to further map banner sizes in the adapter—that'd just restrict the pool of available sizes for Google to fill.

1008k:
@chauduyphanvu Thank you for your confirmation.
I know that AdSize defined in MoPub is passed to AdMob, but in order to realize AdaptiveBanner, you have to calculate the height so that the aspect is correct and create MoPubView dynamically as follows. (I'm not sure if this works properly)
BannerHeightDP = DeviceWidthDP / 320 * 50;

However, in that case, networks that do not support Adaptive Banner, including MoPub, will create unnecessary space in the inventory, which I think is not appropriate.
So I would be happy if the adapter side supports it, or if a similar function is implemented on the MoPub side.

Best Regards.

chauduyphanvu:
@1008k, thank you for your input. 

Our current logic is to let Google know the size, and they will find the best creative that maximizes every pixel of the available ad slot. That should be Google's responsibility and not the adapter's. 

If needed, publishers can later resize the `MoPubView` in the `onAdLoaded()` callback to fit the ad and reduce unnecessary surrounding space. Having said that, if you see any instances of adaptive banner not working as expected, do let us know (with the ad response, a screenshot, and some device logs). Thank you!

PR #310, opened 2020-12-04
[Chartboost] Release 8.2.0 and adapter fix

stosiarz:
We discovered that Cahrtboost adapter is calling initialisation twice which leads to doubled config and install calls. I removed initialisation code from the ChartboostAdapterConfiguration and now every initialisation is done via CharboostShared class which now only allows to initialise SDK if it wasn't previously  (Chartboost.isSdkStarted())

I also planned to publish here update for the sdk 8.2.0 but you already updated.


Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
```diff
@@ -103,18 +101,16 @@ public static synchronized boolean initializeSdk(@NonNull Context context,
         final String appId = serverExtras.get(APP_ID_KEY);
         final String appSignature = serverExtras.get(APP_SIGNATURE_KEY);
 
-        if (!TextUtils.isEmpty(appId) && !TextUtils.isEmpty(appSignature)) {
```
hmertkahraman, line 106: Why did you removed the check for 
```
if appId and appSignature is same as used to initialize CB before, or not
```
Wouldn’t it be useful to prevent calling initialization method again if same appIds were used etc?

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
```diff
@@ -103,18 +101,16 @@ public static synchronized boolean initializeSdk(@NonNull Context context,
...
-            if (appId.equals(mAppId) && appSignature.equals(mAppSignature)) {
-                // We don't need to reinitialize.
-                return false;
-            }
+        if(appId.isEmpty() || appSignature.isEmpty()) {
+            MoPubLog.log(CUSTOM, ADAPTER_NAME, "Chartboost's initialization " +
```
hmertkahraman, line 105: We consider:
```
Chartboost initialization is success if and only if startWithAppId() is called.
```
We’re unsure about the log statement you mention here.

Is calling `startWithAppId` not a guarantee for considering initializing Chartboost?

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
```diff
@@ -52,16 +51,15 @@
...
-    @Nullable
-    private static String mAppSignature;
-
     /**
      * Initialize the Chartboost SDK for the provided application id and app signature.
      */
     public static synchronized boolean initializeSdk(@NonNull Context context,
                                                      @NonNull Map<String, String> serverExtras) {
+        if(Chartboost.isSdkStarted()) {
+            return false;
```
hmertkahraman, line 60: Shouldn’t we return `true` here instead?

How does isSdkStarted really work? We’d ideally should call it all around the ad requests to see if initialization should called or not.

hmertkahraman:
Also, as a side question: Should a CB banner ad request fail when the Location Id parameter is empty (see [here](https://github.com/mopub/mopub-android-mediation/blob/master/Chartboost/src/main/java/com/mopub/mobileads/ChartboostBanner.java#L93-L95)).

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
stosiarz: Actually based on the previous logic
```
if (appId.equals(mAppId) && appSignature.equals(mAppSignature)) {
     // We don't need to reinitialize.
     return false;
}
```
I this we should return false here. When isSDKStarted is true we need to return false as per comment 
" // We don't need to reinitialize."

How does isSdkStarted really work? It returns false until "startWithAppId" is called and sdk was initialised with success meaning delegate didInitialize was called.  

Adding this check before ad request seems like a good idea.

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
stosiarz: This is legacy log taken from the ChartboostAdapterConfiguration class. In current context needs to be rephrased. We just need to indicate here that appId or appSignature is empty which means there were not correctly set in the dashboard. Let me change the text of this log.

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
stosiarz: On the other hand when we do the interstitial cache for example and sdk was not init then SDK will return SESSION_NOT_STARTED error and probably thats better way than to check isSDKStarted before trying to cache and show.

stosiarz:
> Also, as a side question: Should a CB banner ad request fail when the Location Id parameter is empty (see [here](https://github.com/mopub/mopub-android-mediation/blob/master/Chartboost/src/main/java/com/mopub/mobileads/ChartboostBanner.java#L93-L95)).

Banner location should never be null.
```
@NonNull
private String mLocation = ChartboostShared.LOCATION_DEFAULT;
```
But if by any chance this occurs, banner should call fail delegate.

Chartboost/src/main/java/com/mopub/mobileads/ChartboostShared.java
hmertkahraman: Got it, yes the check is great to have. The only thing is the return of this initializeSdk method should return `true` to indicate initialization is complete (was already done), without any failure.

hmertkahraman:
> > Also, as a side question: Should a CB banner ad request fail when the Location Id parameter is empty (see [here](https://github.com/mopub/mopub-android-mediation/blob/master/Chartboost/src/main/java/com/mopub/mobileads/ChartboostBanner.java#L93-L95)).
> 
> Banner location should never be null.
> 
> ```
> @NonNull
> private String mLocation = ChartboostShared.LOCATION_DEFAULT;
> ```
> 
> But if by any chance this occurs, banner should call fail delegate.

Thank you. And if it's `LOCATION_DEFAULT`, how should the ad request respond?

stosiarz:
> LOCATION_DEFAULT

In Case 

> > > Also, as a side question: Should a CB banner ad request fail when the Location Id parameter is empty (see [here](https://github.com/mopub/mopub-android-mediation/blob/master/Chartboost/src/main/java/com/mopub/mobileads/ChartboostBanner.java#L93-L95)).
> > 
> > 
> > Banner location should never be null.
> > ```
> > @NonNull
> > private String mLocation = ChartboostShared.LOCATION_DEFAULT;
> > ```
> > 
> > 
> > But if by any chance this occurs, banner should call fail delegate.
> 
> Thank you. And if it's `LOCATION_DEFAULT`, how should the ad request respond?

In case of `LOCATION_DEFAULT` we should display banner ad as usual

hmertkahraman:
Internally released [here](https://github.com/mopub/mopub-android-mediation/commit/df2f6650f01e78b58663856f81a0af36e3cc9767).

PR #319, opened 2021-01-20
Sdk 2577 handler sdk 4.4.0

gi-adc:
AdColony SDK 4.4
Removed deprecated Handler constructor
Refactoring


hmertkahraman:
This was released internally, please see [here](https://github.com/mopub/mopub-android-mediation/commit/e753b8e6ae062f88fbad6f730398ee6c238d7759).
Noting that we didn't go ahead with the refactoring because of several reasons, but most importantly our SDK's API for each adapter or execution of methods such as `checkAndInitializeSdk ` might be due to change.

PR #322, opened 2021-01-24
Pangle - 3.4.1.1.1

Clyde19-z:
fixed some issues


Pangle/src/main/java/com/mopub/mobileads/PangleAdapterConfiguration.java
```diff
@@ -157,13 +160,30 @@ public static void pangleSdkInit(Context context, String appId) {
...
+    private static JSONArray getAdCallSource() {
+        JSONArray adCallSource = new JSONArray();
+        try {
+            JSONObject mediationObject = new JSONObject();
+            mediationObject.putOpt("name","mediation");
+            mediationObject.putOpt("value","mopub");
+            adCallSource.put(mediationObject);
+            JSONObject adapterVersionObject = new JSONObject();
+            adapterVersionObject.putOpt("name","adapter_version");
+            adapterVersionObject.putOpt("value","1.2.0");
```
p158276, line 180: What is the `adatper_version` related to? Is it a static version number? Any API we can call instead of the static value `1.2.0`?

Clyde19-z: this 'version' represent Pangle's internal code version for Mopub adapter, which will help us to know whether the publisher is using mopub and what function they support. Each time Pangle upgrade the new adapter(with new ad format support or other bug fix), we will change the version number manually.

p158276 (approved):

PR #328, opened 2021-02-19
Google 19.7.0.0

joshuajplagonera:
Updated to use new Native APIs


ericleich:
Only reviewed native here (other code looks like from @chauduyphanvu's branch, but native code LGTM

PR #359, opened 2021-06-19
Icon is not a required field for Google Ad Manager Native Ad

iForests:
As you can see in the [Google Ad Manager’s documentation](https://support.google.com/admanager/answer/6075370?hl=en), Google Ad Manager has two types of Native Ad: Content and App install. For content ads, the "Advertiser logo" is NOT a required field (Always included? = No).

We can display native ads filled from Google Ad Manager (even if it does not have an icon), but the adapter misjudged and gave up the opportunity and revenue. This pull request fixes it by making the icon optional.


chauduyphanvu:
@ericleich, could you take a look at this suggested change? Thanks.

ericleich:
@chauduyphanvu yes this change makes sense to me and is preferred to not filter out the ads, since not all ads have an icon.

chauduyphanvu (approved):

PR #365, opened 2021-07-06
[InMobi] Support for native ad format

RajSuvariya-InMobi:


InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
```diff
@@ -0,0 +1,327 @@
...
+            override fun onAdImpressed(inMobiNative: InMobiNative) {
+                super.onAdImpressed(inMobiNative)
+                MoPubLog.log(AdapterLogEvent.CUSTOM, TAG, "InMobi Native ad is displayed")
+                if (!mIsImpressionRecorded) {
+                    mIsImpressionRecorded = true
+                    notifyAdImpressed()
+                }
+            }
+
+            override fun onAdClicked(inMobiNative: InMobiNative) {
```
p158276, line 232: If the user clicks the same ad multiple times, Is InMobi server end only count 1 time? 
If not we shouldn't do the dedupe logic in the adapter.

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
```diff
@@ -0,0 +1,327 @@
...
+                MoPubLog.log(AdapterLogEvent.CUSTOM, TAG, "InMobi Native ad is clicked")
+                if (!mIsClickRecorded) {
+                    notifyAdClicked()
+                    mIsClickRecorded = true
+                }
+            }
+
+            override fun onAdStatusChanged(inMobiNative: InMobiNative) {
+                super.onAdStatusChanged(inMobiNative)
+                MoPubLog.log(AdapterLogEvent.CUSTOM, TAG, "InMobi Native ad onAdStatusChanged")
```
p158276, line 243: What kind of status change? Are there any detail information that we can address in the log?

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
```diff
@@ -0,0 +1,327 @@
...
+import com.inmobi.ads.listeners.VideoEventListener
+import com.mopub.common.DataKeys
+import com.mopub.common.logging.MoPubLog
+import com.mopub.common.logging.MoPubLog.AdapterLogEvent
+import com.mopub.mobileads.InMobiAdapterConfiguration
+import java.util.*
+
+class InMobiNative : CustomEventNative() {
+
+    private var inMobiStaticNativeAd: InMobiNativeAd? = null
```
p158276, line 23: As you'll set the videoListener [here](https://github.com/mopub/mopub-android-mediation/pull/365/files#diff-806afec5efca112cb5dbe80d2e2aa57e3406669c05da518824a81cf576ca8628R252). To avoid confusion, should we name the variable other than static?

p158276 (changes requested):

p158276:
Hi @RajSuvariya-InMobi ,
Before changing the code, can you merge fetch some log change I put in this repository? (https://github.com/mopub/mopub-android-mediation/tree/inmobi-native-ad)
Thanks.

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
harshit-inmobi: renamed variable to `inMobiNativeAd`

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
harshit-inmobi: Removed this callback method as it is not functional

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
harshit-inmobi: InMobi server does the dedupe for click. we can keep this check in adapter code.

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
```diff
@@ -0,0 +1,327 @@
...
+            ) {
+                super.onAdLoadFailed(inMobiNative, inMobiAdRequestStatus)
+
+                val errorMessage = "InMobi Native request failed " +
+                        "with message: ${inMobiAdRequestStatus.message} " +
+                        "and status code: ${inMobiAdRequestStatus.statusCode}."
+                val nativeErrorCode: NativeErrorCode
+
+                when (inMobiAdRequestStatus.statusCode) {
+                    InMobiAdRequestStatus.StatusCode.INTERNAL_ERROR -> {
```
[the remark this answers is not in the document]
harshit-inmobi: These mappings are specific native ads and are not used anywhere else, so we can keep it this way.

InMobi/src/main/java/com/mopub/nativeads/InMobiNative.kt
harshit-inmobi: @p158276 If I removed the dedupe logic from adapter, will Mopub server count it as multiple clicks and might cause sync issue with InMobi server for number of clicks?

p158276: Hi, we'll only send one time to our server. So you don't need to do the dedupe logic in our adapter.

harshit-inmobi: @p158276  this is fixed, please check

PR #397, opened 2021-09-17
Support for InMobi GDPR

harshit-inmobi:


p158276:
Hi @harshit-inmobi , we just found that GDPR status is updated every time before ad request in iOS. But in Android only update the GDPR flag when initializing. Both ways are ok with us, just note that if the user changes the GDPR flag during the app session you may not able to identify it in Android as it only fetches the GDPR flag during initalize.

harshit-inmobi:
@p158276 In Android we are initializing sdk before every Ad load so GDPR flag will also be updated before every Ad load.

harshit-inmobi:
> Hi @harshit-inmobi , we just found that GDPR status is updated every time before ad request in iOS. But in Android only update the GDPR flag when initializing. Both ways are ok with us, just note that if the user changes the GDPR flag during the app session you may not able to identify it in Android as it only fetches the GDPR flag during initalize.

@p158276 In Android we are initializing sdk before every Ad load so GDPR flag will also be updated before every Ad load.

InMobi/src/main/java/com/mopub/mobileads/InMobiAdapterConfiguration.kt
```diff
@@ -23,6 +28,7 @@ class InMobiAdapterConfiguration : BaseAdapterConfiguration() {
     }
 
     override fun getBiddingToken(context: Context): String? {
+        InMobiSdk.setPartnerGDPRConsent(getGdprConsentObj())
```
p158276, line 31: The `getBiddingToken` triggered when you made the ad request. As you initialise InMobi every time when ad request happened, you don't need to call it again here.

harshit-inmobi: `InMobi.getToken()` will get called before Ad load, in that case the token won't include GDPR info as we'll set GDPR info only after Ad load method is called in Adapter. That is why we are setting GDPR info in `getBiddingToken() `method.
@p158276

p158276: It is more clear if we directly call it right before the ad request, rather than adding it to the getBiddingToken: and trusting that it gets called. Functionally, the method isn't designed for calling GDPR code. If the method's implementation changes in the future, it will break this. Besides, during the initialization you'll also [update the GDPR status.
](https://github.com/mopub/mopub-android-mediation/blob/1d5139a20e814954eba9572757bd20b3f4a7bd90/InMobi/src/main/java/com/mopub/mobileads/InMobiAdapterConfiguration.kt#L95). 
You can also get the GDPR status through the bid request we send to your server.
https://developers.mopub.com/bidders/integration/openrtb_2.5/#3231-object-regsext

kshtgarg21: What exactly do you mean by "directly call it right before the ad request"
what is the exact line you are referring to over here?

I get that bidding token method isn't designed for calling GDPR code

But requirement here is to have updated gdpr consent as a part every token, hence we are calling it before get token.

p158276: Hi @kshtgarg21 ,
After checking with the team, if your SDK need the GDPR consent before generating the token. It totally fine. 
 You can call it before getting the bid token. So let's leave it there.

InMobi/src/main/java/com/mopub/mobileads/InMobiAdapterConfiguration.kt
```diff
@@ -86,6 +92,7 @@ class InMobiAdapterConfiguration : BaseAdapterConfiguration() {
                         inMobiInitCompletionListener?.onFailure(error)
                     }
                 }
+                InMobiSdk.setPartnerGDPRConsent(getGdprConsentObj())
```
p158276, line 95: If you set the GDPR consent here. It will happen after ad load to your SDK as onSuccess happened earlier than this.

We suggest disconnecting the consent setting from the initialization code and call the consent logic before request to ensure consent is set irrespective of the init changes.

Our recommendation should be to set consent status before ad requests. So your SDK can collect GDPR status before request  and aware of consent changes before a traditional ad request

p158276:
Apart from this PR, can I know why Inmobi initialize is called for every ad request? It may increase the latency.

p158276 (changes requested):

harshit-inmobi:
> Apart from this PR, can I know why Inmobi initialize is called for every ad request? It may increase the latency.

@p158276 We are initializing Inmobi sdk for every ad request but internally we have applied checks to not return control if sdk is already initialized. So this would avoid any latency.

InMobi/src/main/java/com/mopub/mobileads/InMobiAdapterConfiguration.kt
harshit-inmobi: this is done, now setting Gdpr before ad request.

rsatrasala (approved):

p158276:
Hi @harshit-inmobi ,
The changes have been reflected in this commit for the release. Close the PR here.
https://github.com/mopub/mopub-android-mediation/commit/74d99bf728f9e15cfa208b77ab6d184c7b6056e3

